PR #15, opened 2021-09-25
Beats time app + watch

tahnok:
Use BMP280 code as a base, but stripped down to bare bones.

Then added a new 'mode' for the current Swatch Internet Time (aka beats)

Been using it for 10 days now with no issues, and it appears to be stable and (usually) accurate


joeycastillo:
yay!

PR #23, opened 2021-11-23
Add initial TOTP watch face impl

tahnok:
Vendor code from https://github.com/Netthaw/TOTP-MCU to do the
heavy lifting of computing SHA-1 and HMAC and the rest of TOTP

Also implement a date_time to unix timestamp method

### DIFF
--- a/movement/lib/TOTP-MCU/LICENSE
+++ b/movement/lib/TOTP-MCU/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 Weravech
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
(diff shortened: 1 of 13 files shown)


tahnok:
A few things that are probably worth discussing here:

 1. Is vendoring the code from https://github.com/Netthaw/TOTP-MCU better than a submodule? I think probably yes for simplicity, unless more submodules get added
 2. Do you want to pull the `current_unix_time_from_rtc` code into a time library?
 3. I think it would make sense to add a UTC offset to the settings (it's used by the beats face as well)
 4. It might make sense to think about some kind of config / settings file that is compiled into the watch for injecting the secret, but would also be nice for setting locations or other complex data that isn't going to change much? Maybe just worth keeping an eye out for other things that could benefit from some compile-time editable settings

joeycastillo:
1. I'm sold on using it as a submodule, as long as we can figure out the additions to the Makefile that would allow submodules to be pulled in automatically when building. 
2. Definitely, and in fact I may get that done today! I'll likely change the function name though just to namespace it.
3. Hard agree; my challenge has been figuring out some of the weird edge cases (like offsets that are 5:30 or 5:45 off from UTC). My current thinking is to have a const global array of the 38 existing time zone offsets, and to dedicate six bits of the movement settings struct to an index within that array. (nobody should need more than 64 time zones)
4. This does make sense; in fact I just had to deal with a similar thing in compiling in a thermistor configuration. While this is a bit further out, we can configure our linker file to reserve an EEPROM emulation area at the end of flash (sized from 256-16384 bytes). This would allow us to stash configuration data in a spot that survives being unplugged and stuffed into the watch case. In a conversation with @wormyrocks a few weeks back, he suggested using a WebUSB interface to write configuration data when plugged in to USB. Definitely further out, but I like that idea as a way to configure these more complex settings.

joeycastillo:
I tested your UNIX time conversion and it looks good! I would definitely like to put this functionality in the watch library. Rather than me pasting the code in, I figure it's preferable for you to add it, to preserve authorship? I added documentation in watch_utility.h and a stub in watch_utility.c; you can just move your function over there (the parameters are all the same).

I also found if you re-add the `timestamp -= utc_offset` line that you had alluded to in an earlier revision, it seems to handle the UTC offsetting quite nicely!

tahnok:
Cool, I've filled out the stub function, thanks for thinking about preserving authorship 😄 

I also properly vendored the TOTP-MCU dependency, I had added it as a submodule and attempted to revert that, but I messed it up. I think it can be added back as a submodule that gets pulled automatically as part of the Makefile, but I'm too tired and new to make to do that well right now. Maybe in a follow up PR?

tahnok:
> This does make sense; in fact I just had to deal with a similar thing in compiling in a thermistor configuration. While this is a bit further out, we can configure our linker file to reserve an EEPROM emulation area at the end of flash (sized from 256-16384 bytes). This would allow us to stash configuration data in a spot that survives being unplugged and stuffed into the watch case. In a conversation with @wormyrocks a few weeks back, he suggested using a WebUSB interface to write configuration data when plugged in to USB. Definitely further out, but I like that idea as a way to configure these more complex settings.

That sounds cool, I wonder if you could have the initial version just be an interactive setup thingy over USB serial, like how the buspirate is configured?

tahnok:
Ok, marked this as ready to merge. I can follow up to have support for multiple codes, since (IMO) small PRs are good PRs

tahnok:
I've also just noticed a bug where I am not left-padding the code with
zeroes. Will fix later tonight unless you get to it first

joeycastillo:
I can make that change! I also need to make a couple of small tweaks to the TOTP library; need to move some of sha1's declarations from the header to the source file, because it's causing linker issues with my version of GCC.

PR #25, opened 2021-12-05
Add tinyusb as git submodules

### LINKED ISSUE #22: MIgrate hathach/tinyusb to a Git submodule
Having the [hathach/tinyusb](https://github.com/hathach/tinyusb) code checked out at this repo means you miss bug fixes and future features, or the burden of constantly updating the code.

I suggest removing the static code and adding back as a [Git submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules), this is the proper way to add other repositories as dependencies to your code. Updates would be done semi-automatically and new clones would always bring the newest version.

### References
- https://github.blog/2016-02-01-working-with-submodules/
- https://www.atlassian.com/git/tutorials/git-submodule

willianpaixao:
Fixes #22


rules.mk
```diff
@@ -20,6 +22,9 @@ $(BUILD)/$(BIN).uf2: $(BUILD)/$(BIN).bin
 	@echo UF2CONV $@
 	@$(UF2) $^ -co $@
 
+$(SUBMODULES):
+	git submodule update --init --recursive
```
joeycastillo, line 26: Removing the `--recursive` flag dramatically cuts down on the time required to pull down submodules (from minutes to seconds), and TinyUSB's submodules all seem to be about building its own tests and examples.

joeycastillo (changes requested):
This is great, exactly what I'd hoped we could do in #22! 

One note: I'd like to remove the `--recursive` flag, at least until we encounter a scenario where we actually need it for a submodule to work.

PR #29, opened 2022-01-01
Countdown

tahnok:
Adds a Countdown face using the new `movement_schedule_background_task` method. I had to add it's opposite method though as well in the form of `movement_cancel_background_task` 

I limited it to 99 minutes and 59 seconds to ensure the watch isn't staying out of deep sleep for too long. I think there's room for a long term? range? countdown that is limited to hours and minutes and just responds to the low energy update event

### DIFF
--- a/movement/watch_faces/complications/countdown_face.h
+++ b/movement/watch_faces/complications/countdown_face.h
@@ -1,0 +1,69 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2022 Wesley Ellis
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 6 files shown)


movement/watch_faces/complications/countdown_face.h
```diff
@@ -0,0 +1,43 @@
...
+Max countdown is 99 minutes and 59 seconds since we have to prevent the watch
+from going to deep sleep using movement_schedule_background_task
+*/
+
+
+typedef enum {
+    cd_waiting,
+    cd_running,
+    cd_setting
+} countdown_mode_t;
```
tahnok, line 18: Is there a way to have a static enum so that these don't contaminate the global namespace? Or maybe I'm just being too paranoid

joeycastillo:
Sorry for the delay in reviewing this! I think it's great. I may have a few improvements in mind, but nothing that blocks this from being merged. The thoughts I have:

* We can check the low energy timeout interval in settings to determine the maximum time; for a 1 hour timeout, the max would be 59:59, for longer ones you could lock it to 99:59 or optionally add hours to the mix.
* I like the idea of holding the alarm button to reset the selected digit... like, if I wanted to set a 3 minute timer but accidentally set it to 4, an alarm long press could reset it to 0 instead of having to punch through to 99 and back.

The only thing I would like to do before merging: would you mind adding the MIT license block at the top of your files? I know I need to add it to some of my watch faces as well; just want to make sure all contributions are explicitly open sourced. You can copy it from main.c and just replace the `Copyright (c) 2021 Joey Castillo` with your own name and date.

Thanks!

tahnok:
Took a while, but I've added the copyright headers to the files

I have been dog-fooding the countdown watch and had also come up a similar idea of being able to reset the current digit with a long press, maybe even having another combo of buttons to decrement by one as I keep going 1 too far and wishing I had a way to go back

movement/watch_faces/complications/countdown_face.h
joeycastillo: I think as long as they're reasonably prefixed, it's nothing to worry about. My instinct would be to do something like `countdown_mode_waiting` here, but the `cd_` prefix seems fine!

joeycastillo:
Awesome! I'm merging this in now; any improvements, we can add in a separate pull request later :)

PR #36, opened 2022-01-19
Move out of settings when face is changed

davidskeck:
While using the Countdown Face, I noticed that I would sometimes change the amount of time to countdown from instead of starting the timer. This occurred despite the fact that the numbers were not blinking.

I believe this is due to leaving the settings mode and entering another watch face. The countdown face remains in settings mode, but since the movement tick frequency reverts back to 1, the blink no longer occurs.

This commit moves into waiting mode if the Countdown Face is in settings mode and the user moves to the next face.

The order of operations to reproduce this on the previous versions is as follows.

Move to Countdown Face -> Enter Settings Mode -> Move out of Countdown Face -> Move back to Countdown Face -> Attempt to start the timer with Alarm button, but the timer amount increases instead.

I tested this commit and it does fix the issue for me.

### DIFF
--- a/movement/watch_faces/complications/countdown_face.c
+++ b/movement/watch_faces/complications/countdown_face.c
@@ -158,6 +158,10 @@ bool countdown_face_loop(movement_event_t event, movement_settings_t *settings, 
             draw(state, event.subsecond);
             break;
         case EVENT_MODE_BUTTON_UP:
+            if (state->mode == cd_setting) {
+                state->mode = cd_waiting;
+                movement_request_tick_frequency(1);
+            }
             movement_move_to_next_face();
             break;
         case EVENT_LIGHT_BUTTON_UP:


davidskeck:
Alternatively, we could set `movement_request_tick_frequency(4);` in `countdown_face_activate` when `state->mode == cd_setting`. This way, the setting state would be retained, but the face would still blink. I think both approaches are acceptable. Any thoughts?

joeycastillo:
Good catch, thanks! My instinct had been that watch faces should always call `movement_request_tick_frequency(1)` in their `resign` function so that it resets before the next watch face takes over. But thinking about it now, I'm wondering if that should be the responsibility of the Movement framework itself. 

I am leaning toward making this change one level up, in movement.c's `app_loop`, to prevent watch faces from having to manage this themselves.

davidskeck:
Very interesting. So would Movement manage the context of the tick frequency for each face? Like reapply the 4Hz tick to Countdown when it is reselected later?

joeycastillo:
I think we'd just formalize / document the fact that when your watch face comes on screen, you're getting a 1 Hz tick, and if you need something other than that you should set it in your watch face's `activate` function. (unrelatedly, I also think it makes sense to leave 'setting' mode when going off screen / before coming on screen, which would prevent countdown from needing a 4 Hz tick when it is activated)

tahnok:
Ya, I didn't intended to stay in settings mode when you leave the face, so
when resigning it would be good to change the state to waiting if we are in
setting.

100% agree that movement should make sure the tick rate gets reset whenever
a new face is activated. #banfootguns

PR #37, opened 2022-01-19
Countdown: Added ability to reset to default timer value

davidskeck:
Added option to reset to the default timer value when in setting mode by long press of alarm. This was useful to me as I sometimes accidentally increase the timer value and then would need to loop around to get back to the default value.


joeycastillo:
Just tested and this works great, thanks for the PR!

PR #38, opened 2022-01-19
Simple Clock: Update to signal exactly on the hour

davidskeck:
I was getting the buzzer at 59 minutes past the hour, not exactly on the hour. This changes to buzzing at exactly HOUR:00:00. Tested this locally.


joeycastillo:
Tested here too, looks good!

PR #39, opened 2022-01-19
Reset selection to 0 on resign from setting state

davidskeck:
I believe this also needs to be set back to 0 so the setting screen will resume at minutes, not potentially at settings on next run.


davidskeck:
Sorry to make a lot of small PRs. Let me know if there is another way you'd like me to do these changes! Having a blast :)

joeycastillo:
I can roll with small pull requests, thanks!

PR #43, opened 2022-01-25
Add new function to disable certain RTC periodic callbacks

a2:
I added a `watch_rtc_disable_matching_periodic_callbacks()` function to remove the explicit Emscripten checks in `movement/movement.c`. This function takes a bit mask of callbacks to disable. For example, to disable the 2 Hz and 4 Hz callbacks, pass a `uint8_t` of `0b01100000`. Note that this is not the same as `2 | 4`. This bit mask encoding works because only powers of 2 are supported.

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -155,13 +155,7 @@ void movement_request_tick_frequency(uint8_t freq) {
     if (freq == 128) return; // Movement uses the 128 Hz tick internally
 
     // disable all callbacks except the 128 Hz one
-#if __EMSCRIPTEN__
-    for (int i = 1; i < 128; i = i << 1) {
-        watch_rtc_disable_periodic_callback(i);
-    }
-#else
-    RTC->MODE2.INTENCLR.reg = 0xFE;
-#endif
+    watch_rtc_disable_matching_periodic_callbacks(0b01111111);
 
     movement_state.subsecond = 0;
     movement_state.tick_frequency = freq;
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -155,7 +155,7 @@ void movement_request_tick_frequency(uint8_t freq) {
     if (freq == 128) return; // Movement uses the 128 Hz tick internally
 
     // disable all callbacks except the 128 Hz one
-    watch_rtc_disable_matching_periodic_callbacks(0b01111111);
+    watch_rtc_disable_matching_periodic_callbacks(0xFE);
 
     movement_state.subsecond = 0;
     movement_state.tick_frequency = freq;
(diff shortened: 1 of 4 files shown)


movement/movement.c
```diff
@@ -155,13 +155,7 @@ void movement_request_tick_frequency(uint8_t freq) {
...
 
     // disable all callbacks except the 128 Hz one
-#if __EMSCRIPTEN__
-    for (int i = 1; i < 128; i = i << 1) {
-        watch_rtc_disable_periodic_callback(i);
-    }
-#else
-    RTC->MODE2.INTENCLR.reg = 0xFE;
-#endif
+    watch_rtc_disable_matching_periodic_callbacks(0b01111111);
```
joeycastillo, line 158: aha! I think this needs to be 0b11111110. Was getting some funky behavior in the beats clock, it's backwards from what you expect but the low bit is the high frequency tick, and the high bit is the 1 Hz tick.

a2: Ah that makes sense. I wonder if I should modify the implementation to accept this "intent", though? It makes more sense to me this way at least.

joeycastillo: I dunno, the joy of this one is that it's one line on the hardware, `RTC->MODE2.INTENCLR.reg = mask`; if we mirror it to make it feel right, that'll end up requiring additional bit twiddling on the hardware side. It's also used rarely enough that as long as we document it, I think it's fine if it's goofy.

joeycastillo:
hmm, there is still something funky happening here, which is funny because when I just changed the value to 0b11111110 earlier without making any other changes, both the device and the simulator were in sync. But with this latest change, it seems like some callbacks are sticking around on the simulator, causing the tick function to get called too often. This screenshot is after cycling through the faces a few times and landing on beats, with logging statements at the end of `cb_tick` and at the beginning of `beats_face_loop`: on the right on device, there are 8 subseconds logged, and a reset when the second changes, whereas on the left the tick function is getting called way more often in the simulator. 

Screen Shot 2022-01-25 at 6 55 09 PM

I sense it's something to do with `emscripten_clear_interval` but I'd need to investigate more.

a2:
One thing I noticed while debugging #45 is that the timer IDs can be `0`, but I believe they cannot be negative. Going to investigate this more tomorrow evening if I have time.

a2:
My plan is to rebase this PR after (hopefully) merging the buzzer changes since I don't want to have multiple PRs that affect the way the timeout/interval/rAF IDs are handled. But in the mean time I'm going to test that rebase locally.

joeycastillo:
yay, looking good over here!

PR #44, opened 2022-01-25
Remove ADC_* defines in favor of hardware-specific mapping

a2:
I removed the pragma defines of `ADC_REFCTRL_REFSEL_INTREF_Val`, `ADC_REFCTRL_REFSEL_INTVCC0_Val`, `ADC_REFCTRL_REFSEL_INTVCC1_Val`, and `ADC_REFCTRL_REFSEL_INTVCC2_Val`. As discussed in simulator PR, these values are not useful to a watch developer and only really apply to the SAM L22 hardware itself.

### DIFF
--- a/watch-library/hardware/watch/watch_adc.c
+++ b/watch-library/hardware/watch/watch_adc.c
@@ -23,6 +23,7 @@
  */
 
 #include "watch_adc.h"
+#include "driver_init.h"
 
 static void _watch_sync_adc(void) {
     while (ADC->SYNCBUSY.reg);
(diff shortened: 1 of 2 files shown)


watch-library/hardware/watch/watch_adc.c
```diff
@@ -138,13 +139,30 @@ void watch_set_analog_sampling_length(uint8_t cycles) {
...
+            return ADC_REFCTRL_REFSEL_INTVCC0_Val;
+            break;
+        case ADC_REFERENCE_VCC_DIV2:
+            return ADC_REFCTRL_REFSEL_INTVCC1_Val;
+            break;
+        case ADC_REFERENCE_VCC:
+            return ADC_REFCTRL_REFSEL_INTVCC2_Val;
+            break;
+    }
+}
```
a2, line 157: This could probably be a lookup table since the enum is an integer and these values are contiguous.

joeycastillo: Yep yep, but for now this is clear — and moving it to an enum matches with how the watch library tends to define other things of this nature.

joeycastillo:
Tested in the sim and on device, looks good!

PR #45, opened 2022-01-25
Implement buzzer methods with AudioContext API

a2:
This PR adds buzzer support for the Sensor Watch simulator. I used the `apps/buzzer-test` app for testing.

### DIFF
--- a/apps/buzzer-test/app.c
+++ b/apps/buzzer-test/app.c
@@ -115,7 +115,7 @@ bool app_loop(void) {
             900,
         };
         application_state.play = false;
-        for(size_t i = 0; i < sizeof(rains); i++) {
+        for(size_t i = 0, count = sizeof(rains) / sizeof(rains[0]); i < count; i++) {
             char buf[9] = {0};
             if (rains[i] == BUZZER_NOTE_REST) {
                 printf("rest for %d ms\n", durations[i]);
--- a/rules.mk
+++ b/rules.mk
@@ -13,6 +13,7 @@ endif
 $(BUILD)/$(BIN).html: $(OBJS)
 	@echo HTML $@
 	@$(CC) $(LDFLAGS) $(OBJS) $(LIBS) -o $@ \
+		-s ASYNCIFY=1 \
 		-s EXPORTED_FUNCTIONS=_main \
 		--shell-file=$(TOP)/watch-library/simulator/shell.html
 
--- a/watch-library/simulator/watch/watch_deepsleep.c
+++ b/watch-library/simulator/watch/watch_deepsleep.c
@@ -35,6 +35,7 @@ static uint32_t watch_backup_data[8];
 
 void watch_register_extwake_callback(uint8_t pin, ext_irq_cb_t callback, bool level) {
     if (pin == BTN_ALARM) {
+        watch_enable_external_interrupts();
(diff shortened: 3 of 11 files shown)


apps/buzzer-test/app.c
```diff
@@ -115,7 +115,7 @@ bool app_loop(void) {
             900,
         };
         application_state.play = false;
-        for(size_t i = 0; i < sizeof(rains); i++) {
+        for(size_t i = 0, count = sizeof(rains) / sizeof(rains[0]); i < count; i++) {
```
a2, line 118: I think this worked before because the ARM C compiler treated each note enum as an `int8_t` due to its size. From my testing, the Emscripten C compiler behaves differently.

rules.mk
```diff
@@ -13,6 +13,7 @@ endif
 $(BUILD)/$(BIN).html: $(OBJS)
 	@echo HTML $@
 	@$(CC) $(LDFLAGS) $(OBJS) $(LIBS) -o $@ \
+		-s ASYNCIFY=1 \
```
a2, line 16: This adds support for the `emscripten_sleep()` function.

watch-library/simulator/watch/watch_deepsleep.c
```diff
@@ -35,6 +35,7 @@ static uint32_t watch_backup_data[8];
 
 void watch_register_extwake_callback(uint8_t pin, ext_irq_cb_t callback, bool level) {
     if (pin == BTN_ALARM) {
+        watch_enable_external_interrupts();
```
a2, line 38: If a watch app explicitly uses `watch_register_extwake_callback()` to wake with the Alarm button, we need to enable interrupts (a simulator implementation detail).

joeycastillo: right because on the hardware, that interrupt is part of the RTC's responsibilities, makes sense!

joeycastillo:
So with this one, I am afraid I'm getting an exception when I press the Mode button. Seems unrelated to the next watch face; it also happens when the setting or voltage face is next in the carousel. Tested in Safari and Firefox.

Screen Shot 2022-01-25 at 6 40 25 PM

joeycastillo:
First half of that stack trace:

Screen Shot 2022-01-25 at 6 43 22 PM

a2:
Spent some interesting debugging time on this one. I _think_ that there is a race condition (which, of course it is) due to [reentrancy](https://emscripten.org/docs/porting/asyncify.html#reentrancy). (Ehe event handlers are still receiving and processing events while the `emscripten_sleep()` call is waiting.) Tomorrow I'll try refactoring the event handling code to be in "pure" (inline) JavaScript again so I can do more to prevent this condition.

a2:
tl;dr: I tested `movement` and `buzzer-test` and both seem to be working even when I spam the A/L/M keys or lick on the SVG buttons. I also updated some of the interval/timeout/rAF IDs to check against `-1` instead of `0` because `0` is a valid "first" ID.

I think the original issue was the simulator code has a few "entrances": the main loop, which is driven by `requestAnimationFrame()`, and the event handlers. If one of those is triggered during `emscripten_sleep()` (the underlying delay mechanism) then (I think) the `Asyncify` state corrupts and the code is not able to "unwind" correctly after `emscripten_sleep()` completes.

joeycastillo:
Tested both the buzzer test app and Movement in Firefox and on hardware, both seem to work great! Merging…

PR #46, opened 2022-01-25
Add GitHub Workflow job to build and archive simulator

a2:
I thought it might be useful to add a workflow job that builds & archives the Movement sub-project for the simulator target. At least to make sure everything works correctly before merging.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,7 +40,6 @@ jobs:
         run: emmake make
         working-directory: 'movement/make'
       - name: Archive simulator build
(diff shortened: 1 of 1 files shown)


joeycastillo:
Love this! Would also love to eventually have a workflow that deploys it to a server. I sense it's as simple as uploading the artifacts and serving them? No fancy requirements?

a2:
> I sense it's as simple as uploading the artifacts and serving them? No fancy requirements?

Exactly, any static site host that can serve the three files inside the archive will work.

Issue #42, opened 2022-01-20
Small updates to Movement README


davidskeck:
Hello,

I would like to suggest a few updates to the Movement README.

1. Consider adding a section that details the need to add the new watch face file to the Makefile.
2. Consider explaining the purpose of the Boolean return in watch_face_loop.

These two items tripped me up when making my own watch face.

Thanks!

joeycastillo:
Thanks for the feedback! Definitely agree on no. 2, I need to do a better job explaining that. 

As for no. 1, I think the next step for movement is to try to automate some of this. Ideally I'd like to have like a Python script that one can run something like `watch-face new tide_chart`, and have it generate the header and implementation file with boilerplate, and handle the appropriate includes and Makefile additions. This would naturally involve a bigger rewrite of the README to accommodate the new process, but I'd like too get started on this soon so I might put a pin in this issue and address all README tasks when that's done.

tahnok:
In the short term / low effort, you could link to a PR that adds a watch
face as an example for people to follow

davidskeck:
> Thanks for the feedback! Definitely agree on no. 2, I need to do a better job explaining that.
> 
> As for no. 1, I think the next step for movement is to try to automate some of this. Ideally I'd like to have like a Python script that one can run something like `watch-face new tide_chart`, and have it generate the header and implementation file with boilerplate, and handle the appropriate includes and Makefile additions. This would naturally involve a bigger rewrite of the README to accommodate the new process, but I'd like too get started on this soon so I might put a pin in this issue and address all README tasks when that's done.

That sounds great. Would you like any help making that script? Agree that linking to a PR short term would be helpful.

joeycastillo:
> Would you like any help making that script?

I would welcome that help! I sense I would get to this eventually but if it's something you're interested in, by all means, I'd be stoked if someone wanted to take a crack at it!

The conversation around this issue lit the fire under me to [add a template](https://github.com/joeycastillo/Sensor-Watch/commit/f819a18791d6ad5a4f68125928587d0f87833a1c) and slightly reorganize the way we're doing watch faces (moved the lengthy list of `#include`s to a separate file, and renamed a couple of folders). Realizing now that the folder structure means we need two parameters, the type of watch face and the name (i.e. `watch-face new complication tide_chart` or `watch-face new clock mars_clock`. 

[...]

PR #47, opened 2022-01-27
Added rough draft of watch_face script 

davidskeck:
Added script to automatically generate new watch faces from the command line.
Updated template file to add missing includes and correct the placeholder in memset.

Let me know if this is along the lines of what you were thinking in #42.

I'd love any suggestions to improve this code as well. Thank you!

### DIFF
--- a/movement/template/template.c
+++ b/movement/template/template.c
@@ -78,7 +78,7 @@ bool <#watch_face_name#>_face_loop(movement_event_t event, movement_settings_t *
             break;
     }
 
-    // return true if the watch can enter standby mode. If you are PWM'ing an LED or buzzing the buzzer nere,
+    // return true if the watch can enter standby mode. If you are PWM'ing an LED or buzzing the buzzer here,
     // you should return false since the PWM driver does not operate in standby mode.
     return true;
 }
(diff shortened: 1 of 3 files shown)


davidskeck:
Some context for the adding of a newline to the template files: https://stackoverflow.com/questions/5813311/whats-the-significance-of-the-no-newline-at-end-of-file-log

joeycastillo:
This sparks so much joy. Thank you for doing this! I don't have any significant comments; just tested it here, and it works great. 😄

joeycastillo:
Actually, one thing occurred to me just after merging: we can autopopulate the year, it doesn't need to be a command line argument. I just went ahead and [made the change here](https://github.com/joeycastillo/Sensor-Watch/commit/5d5e5b125a383174f1891d37498415cc10fe84fe). 

We could also probably autopopulate the author by parsing `git config --list`, but that seems a bit creepy and probably has some privacy implications. 🙃

davidskeck:
Excellent! I'm so glad you found it useful.

I thought of the year thing, but forgot about it when I started implementing things! Good call. I agree on the author thing -- I thought of that too, but decided against it for the same reason.

I have a couple more ideas to improve the script (mainly adding output to say what files were changed/added and their locations), which I'll add in a separate PR.

Thanks for this project! It is super fun.

PR #49, opened 2022-02-05
Added output to watch face script

davidskeck:
changed `--author_name` to `--author-name` to match typical style of command line arguments
changed some function names and implementations to improve readability
changed to `sys.exit` instead of `exit` as `sys.exit` is recommended for use in programs and `exit` is intended for use in the interactive shell (this can cause issues for some terminal emulators)

### DIFF
--- a/movement/template/watch_face.py
+++ b/movement/template/watch_face.py
@@ -22,6 +22,7 @@
 
 import os
 import re
+import sys
 import datetime
 import argparse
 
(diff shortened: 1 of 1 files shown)


davidskeck:
Some [context](https://stackoverflow.com/questions/1253679/should-command-line-options-in-posix-style-operating-systems-be-underscore-style) for the argument name change.
And [context](https://stackoverflow.com/questions/6501121/difference-between-exit-and-sys-exit-in-python) for `sys.exit`, because I found it interesting.

joeycastillo:
Looks good to me!

PR #53, opened 2022-02-06
Make stopwatch work off-screen and in low energy mode

joeycastillo:
Opening this mostly for @tahnok; I made a fair number of changes to your stopwatch face, and wanted to run them by you. I mentioned in #21 that if we used a start time to track the stopwatch duration it could work in the background without relying on the tick callback. I implemented that here. I also reformatted the text a bit to allow days in the top row and hours/minutes/seconds on the bottom, since it can run for up to 40 days now.

There's a demo running in the emulator [at this link](https://joeycastillo.sdf.org/sensor-watch/watch.html).

One limitation is that the displayed time will be incorrect if the wearer resets the system time while the stopwatch is running, but this seems like an edge case that we can document. Or maybe movement's settings struct needs a flag to signal to watch faces that the time has changed since they were last on screen.

### DIFF
--- a/movement/watch_faces/complication/stopwatch_face.c
+++ b/movement/watch_faces/complication/stopwatch_face.c
@@ -1,46 +1,95 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2022 Wesley Ellis
+ * Copyright (c) 2022 Joey Castillo
+ *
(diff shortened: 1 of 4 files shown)


tahnok:
Looks good to me!

PR #54, opened 2022-02-23
Updated function brief

davidskeck:
I think this brief was copied from an earlier function and is not correct for this function.


joeycastillo:
So it was, thanks!

PR #61, opened 2022-03-28
Tomato timer: pomodoro style timer

tahnok:
Add a tomato timer watch face that alternates between 25 and 5 minute timers as in the [pomodoro technique](https://en.wikipedia.org/wiki/Pomodoro_Technique). I think Pomodoro might be copyright or something, so we just call it the tomato timer

image

The top right letter shows mode (f for focus or b for break). The bottom right shows how many focus sessions you've completed. (You can reset the count with a long press of alarm)

- When you show up and it says 25 minutes, you can start it (alarm), switch to 5 minute (light) mode or leave (mode)
- When it's running you can reset (alarm), or leave (mode)
- when it's done, we beep and go back to step 1, changing switching mode from focus to break (or break to focus)

I also pulled out a watch_utility_offset_timestamp and put it in watch_utitlity

### DIFF
--- a/movement/watch_faces/complication/tomato_face.h
+++ b/movement/watch_faces/complication/tomato_face.h
@@ -1,7 +1,7 @@
 /*
  * MIT License
  *
- * Copyright (c) 2022 Joey Castillo
+ * Copyright (c) 2022 Wesley Ellis
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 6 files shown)


movement/watch_faces/complication/tomato_face.h
```diff
@@ -0,0 +1,63 @@
...
+typedef enum {
+    to_ready,
+    to_run,
+    // to_pause, // TODO implement pausing
+} tomato_mode;
+
+typedef enum {
+    to_break,
+    to_focus,
+} tomato_kind;
```
joeycastillo, line 39: It's more verbose, but I would name these `tomato_mode_ready`, `tomato_kind_break`, etc. Keeps the global namespace cleaner, and makes it clearer when assigning or comparing values.

joeycastillo (commented):
This is wonderful, thank you for making this! Folks are going to be stoked for this one.

I have one minor issue, mentioned inline, about the naming of some of the enum values. If you can make that change, I'd love to merge it in.

movement/watch_faces/complication/tomato_face.h
tahnok: Roger, names changed

joeycastillo:
hooray!

PR #66, opened 2022-04-30
Parallelize alternate fw builds

GeorgeHahn:
The alternate fw build script performs a bunch of builds. Parallelizing the compilation of these builds speeds up the process by quite a bit.

I suspect you may have omitted this for a reason. If that's the case, I'd like to switch this PR to add some notes around parallel compilation issues.

### DIFF
--- a/make.mk
+++ b/make.mk
@@ -24,6 +24,7 @@ else
 endif
 $(if ${VERBOSE},$(info OS detected: $(DETECTED_OS)))
 
+ifeq ($(OS), Windows_NT)
   MKDIR = gmkdir
 else
   MKDIR = mkdir
--- a/make.mk
+++ b/make.mk
@@ -9,7 +9,21 @@ endif
 ##############################################################################
 .PHONY: all directory clean size
 
-ifeq ($(OS), Windows_NT)
+# OS detection, adapted from https://gist.github.com/sighingnow/deee806603ec9274fd47
+DETECTED_OS :=
+ifeq ($(OS),Windows_NT)
+  DETECTED_OS = WINDOWS
+else
+  UNAME_S := $(shell uname -s)
+  ifeq ($(UNAME_S),Linux)
+    DETECTED_OS = LINUX
+  endif
+  ifeq ($(UNAME_S),Darwin)
+    DETECTED_OS = OSX
+  endif
+endif
+$(if ${VERBOSE},$(info OS detected: $(DETECTED_OS)))
+
   MKDIR = gmkdir
 else
   MKDIR = mkdir
--- a/make.mk
+++ b/make.mk
@@ -29,6 +29,17 @@ else
   MKDIR = mkdir
 endif
 
+ifeq ($(DETECTED_OS), LINUX)
+  MAKEFLAGS += -j `nproc`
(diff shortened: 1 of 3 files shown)


joeycastillo:
This seems to work great! To be honest I didn't omit this for any particular reason; I just didn't know about the option. 

I am curious though what issues you may have anticipated with parallel compilation, just so I can be on the lookout for them. But at a glance, all the firmware images I've tested compile properly and run fine both on the watch and in the simulator.

GeorgeHahn:
I used to work with an embedded engineer who didn't trust anything but
clean sequential builds 😛. This opens up the potential for race conditions
in the build. I'm far from a make expert and I don't have any particular
examples in mind. Here are some I found on StackOverflow:
https://stackoverflow.com/questions/26511319/parallel-makefile-execution-causes-a-library-to-be-built-twice,
https://stackoverflow.com/questions/37873522/unreliable-parallel-builds-in-a-makefile-with-intermediate

One change that I could see requiring care in this repo is if the makefile
were changed to allow switching between variants without fully cleaning the
build first. (I'm imagining keeping the built watch faces and rebuilding
the rest of the project.) That would need to be implemented carefully to
ensure the partial clean and rebuild are done in the correct order.

I learned something while looking for the examples above: the change I
wrote here may not work on mac os. The flag can also be specified inside
the makefile so it's active for individual make runs. I'd like to look into
this further before merging:
https://stackoverflow.com/questions/51148963/makefile-run-targets-in-parallel

GeorgeHahn:
This should now work on Linux, OSX, and Windows. I tested builds on Linux and verified that the `NUMBER_OF_PROCESSORS` variable exists on Windows. Could you test it on OSX?

joeycastillo:
FWIW, the previous solution (putting it in just the alternate firmware sketch) did work on OSX for me. But now that it's part of the Makefile, the tinyUSB build step fails because it's trying to build tinyUSB before the `git submodule update --init` step is complete. 

We might be able to fix this by moving any submodule compilation to the end of the SRCS list, although a slow network connection could still cause it to fail if the compiler is faster than the network. I'm not sure what make-fu would be required to make sure that submodules are installed before farming out compilation to other cores.

GeorgeHahn:
It looks like there's a clean solution to the ordering problem: order-only dependencies. I used that to instruct make to create the build directory and initialize submodules before building anything.

That solution caused a small problem: `make clean` would run the `git submodule` rule (but only on the first execution).

At first I assumed this was caused by some file named `tinyusb` overlapping with the submodule rule of the same name. The behavior didn't really line up for that though. Still, the submodule rule appears to be a reasonable candidate for a phony rule, so I marked it as one.

In the end, I tracked this down to the wildcard include at the bottom of `rules.mk`. Removing that line caused the submodule invocation to stop, and my builds kept working. I'm not sure if the line is necessary in other scenarios or environments, so I attempted to preserve it by switching it to a more targeted wildcard. It may be possible to remove this line entirely - it's also possible that I broke whatever it was being used for.

joeycastillo:
After testing this for a while it seems to work great, so I'm merging it in. Thanks for the PR!

PR #68, opened 2022-05-10
Optimization: compare raw ADC battery readings

GeorgeHahn:
This is a silly optimization to make. It saves about four instructions
once a day when the battery is checked. There is a tiny bit of extra
hidden savings: those instructions happen while the ADC is enabled.
I'm sure it saves whole picojoules!

Note the operator change from less than to less than or equal. This is
due to a rounding difference when the raw ADC value is exactly equal to
the threshold value. I didn't work this through to a definitive answer
for where the rounding is happening. After the operator change, the code
displays equivalent behavior.

See this godbolt for more details https://godbolt.org/z/9W7vha5P6

### DIFF
--- a/movement/watch_faces/clock/simple_clock_face.c
+++ b/movement/watch_faces/clock/simple_clock_face.c
@@ -72,11 +72,13 @@ bool simple_clock_face_loop(movement_event_t event, movement_settings_t *setting
             if (date_time.unit.day != state->last_battery_check) {
                 state->last_battery_check = date_time.unit.day;
                 watch_enable_adc();
-                uint16_t voltage = watch_get_vcc_voltage();
+            uint32_t voltage = watch_get_raw_vcc_voltage();
                 watch_disable_adc();
                 // 2.2 volts will happen when the battery has maybe 5-10% remaining?
                 // we can refine this later.
-                state->battery_low = (voltage < 2200);
+            uint16_t low_batt_threshold_mv = 2200;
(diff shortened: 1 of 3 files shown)


joeycastillo:
Two notes: 

1. The simulator build failed because the `watch_get_raw_vcc_voltage` function was only implemented in the hardware target, not the simulator.
2. Do you think this will save power significantly enough to justify the decreased code readability? I know it's a few more instructions, but if the tradeoff is that it adds a footgun for someone implementing a battery voltage check, I'm unsure whether it's worth it…

GeorgeHahn:
I noticed that my comparison was using incorrect compiler flags. Here's a better comparison: https://godbolt.org/z/6rPss46bW. Unfortunately I don't have a way to power profile this on hardware, so I can't give any real numbers.

That being said, there's almost no way this makes a meaningful difference for the once a day check. 

Just to be sure, I checked the instruction set summary ([here](https://developer.arm.com/documentation/ddi0484/c/Programmers-Model/Instruction-set-summary?lang=en)) to get the cycle duration of a multiply instruction. A multiply can either take 1 or 32 cycles, depending on the core configuration. The SAM L22 family data sheet clarifies that its core uses the single cycle multiplier implementation (DS60001465B section 10.1.1). So we're looking at 5 cycles in a relatively high power state, once a day. My back of napkin math for the power consumption over a year has so many zeros in it that it's not even worth putting a number to.

I'm going to close this, but I'm totally happy to fix the simulator side and clear the API up with docs, API privatization, or other improvements if you'd like to land it.

PR #69, opened 2022-05-10
Probability watch face

spencerbyw:
### Probability watch face
- Allows you do do random number rolls for 2, 4, 6, 8, 10, 12, 20, and 100 sided dice.
- For a demo video, see https://twitter.com/spencerbyw/status/1518975413849968641
- I feel like I followed the other watch faces as examples as best I could, but happy to make adjustments if there's anything I can improve!


movement/watch_faces/complication/probability_face.c
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+}
+
+
+// ---------------------------
+// Standard watch face methods
+// ---------------------------
+void probability_face_setup(movement_settings_t *settings, uint8_t watch_face_index, void ** context_ptr) {
+    (void) settings;
```
joeycastillo, line 111: This is my fault, I need to add it to the templates, but there should be a `(void) watch_face_index;` line here to silence the warning about an unused parameter.

movement/watch_faces/complication/probability_face.c
```diff
@@ -0,0 +1,181 @@
...
+
+// --------------
+// Custom methods
+// --------------
+
+static void display_dice_roll(probability_state_t *state) {
+    char buf[7]; // Num chars needed + 1 for \0
+    if (state->rolled_value == 0) {
+        if (state->dice_sides == 100) {
+            sprintf(buf, " C    ", state->dice_sides);
```
joeycastillo, line 44: This sprintf doesn't need the `state->dice_sides` parameter since it doesn't have any variables in the format string.

joeycastillo, line 44: This sprint doesn't need the `state->dice_sides` parameter since it doesn't have any variables in the format string.

movement/watch_faces/complication/probability_face.c
```diff
@@ -0,0 +1,181 @@
...
+void probability_face_activate(movement_settings_t *settings, void *context) {
+    (void) settings;
+    probability_state_t *state = (probability_state_t *)context;
+
+    state->dice_sides = DEFAULT_DICE_SIDES;
+    state->rolled_value = 0;
+}
+
+bool probability_face_loop(movement_event_t event, movement_settings_t *settings, void *context) {
+    probability_state_t *state = (probability_state_t *)context;
```
joeycastillo, line 130: Same thing here, a `(void) settings;` line silences the unused parameter warning.

joeycastillo, line 130: Same thing here, a `(void) settings;` line silences the unused parameter warning.

movement/watch_faces/complication/probability_face.h
```diff
@@ -0,0 +1,51 @@
...
+
+#ifndef PROBABILITY_FACE_H_
+#define PROBABILITY_FACE_H_
+
+#include "movement.h"
+
+typedef struct {
+    uint16_t dice_sides;
+    uint16_t rolled_value;
+    uint16_t animation_frame;
```
joeycastillo, line 33: Any reason these couldn't be `uint8_t`s? It would squish the whole state struct down to four bytes, and silence a warning where sprint thinks the buffer could overflow with values >999.

joeycastillo, line 33: Any reason these couldn't be `uint8_t`s? It would squish the whole state struct down to four bytes, and silence a warning where sprint thinks the buffer could overflow with values >999.

movement/watch_faces/complication/probability_face.c
```diff
@@ -0,0 +1,181 @@
...
+// Standard watch face methods
+// ---------------------------
+void probability_face_setup(movement_settings_t *settings, uint8_t watch_face_index, void ** context_ptr) {
+    (void) settings;
+    if (*context_ptr == NULL) {
+        *context_ptr = malloc(sizeof(probability_state_t));
+        memset(*context_ptr, 0, sizeof(probability_state_t));
+
+        // Seed random number generator
+        srand(time(NULL));
```
joeycastillo, line 117: You can remove this line entirely if you switch to using arc4random, since it'll regularly pull entropy from the true hardware RNG.

joeycastillo, line 117: You can actually remove this line entirely, if you switch to using arc4random instead.

movement/watch_faces/complication/probability_face.c
```diff
@@ -0,0 +1,181 @@
...
+// According to SO this is the better way to roll a random number to avoid introducing skew.
+// https://stackoverflow.com/a/2999130
+// (As opposed to using modulus like: `state->rolled_value = rand() % state->dice_sides + 1;`)
+static void generate_random_number(probability_state_t *state) {
+    int limit = state->dice_sides - 1;
+    int divisor = RAND_MAX/(limit+1);
+    int retval;
+
+    do {
+        retval = rand() / divisor;
```
joeycastillo, line 71: Suggest using arc4random here instead (see comments in main review).

joeycastillo, line 71: Suggest using arc4random here instead (see comments in main review).

joeycastillo (changes requested):
Looks great! I have a few comments, mostly on silencing warnings, but I have two real notes, one on style and one on substance. 

On the style side: I think it would be rad to add a two-letter title in the top line, just to make sure the wearer recognizes which watch face this is. Maybe just "PR" up top for "Probability", in the position where Stopwatch has ST and Temperature has TE. 

Second: I realize it's not well documented in the watch library, but if you use `arc4random` instead of `rand`, Sensor Watch will [automatically pull entropy from the SAM L22's hardware random number generator](https://github.com/joeycastillo/Sensor-Watch/blob/main/watch-library/hardware/watch/watch_private.c#L85). This is the highest quality source of entropy on the device, and it eliminates the need to seed the PRNG manually: on first call, `arc4random` will automatically call `getentropy` and seed itself with the resulting random data.

The True Random Number Generator (TRNG) generates unpredictable random numbers that are not generated by an algorithm. It passes the American NIST Special Publication 800-22 and Diehard Random Tests Suites. The TRNG may be used as an entropy source for seeding an NIST approved DRNG (Deterministic RNG) as required by FIPS PUB 140-2 and 140-3.

spencerbyw:
@joeycastillo Thanks for the thorough review! These are all great ideas. I'll work on getting these comments addressed over the weekend.

movement/watch_faces/complication/probability_face.h
spencerbyw: You're right, a uint8_t should work fine, I'll make that change.

movement/watch_faces/complication/probability_face.c
spencerbyw: Just an FYI on this, it seems since I was doing `#include <time.h>` in this file it would result in this error when I tried to make a uf2 build:
```LD build/watch.elf
/opt/gcc-arm-none-eabi/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: /opt/gcc-arm-none-eabi/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/lib/thumb/v6-m/nofp/libc.a(lib_a-gettimeofdayr.o): in function `_gettimeofday_r':
gettimeofdayr.c:(.text._gettimeofday_r+0xe): undefined reference to `_gettimeofday'
collect2: error: ld returned 1 exit status
make: *** [../../rules.mk:24: build/watch.elf] Error 1```

Hoping that removing that include and using arc4random instead should give me the result I want.

**Comment by spencerbyw** [comment]
@joeycastillo I made the changes you requested, however after making the `arc4random` changes it builds fine to UF2, but when I try to build for the emulator I get the following error:
```
CC build/probability_face.o
../watch_faces/complication/probability_face.c:62:27: error: implicit declaration of function 'arc4random_uniform' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
    state->rolled_value = arc4random_uniform(state->dice_sides) + 1;
                          ^
1 error generated.
emcc: error: '/home/linuxbrew/.linuxbrew/Cellar/emscripten/3.1.8/libexec/llvm/bin/clang -target wasm32-unknown-emscripten -DEMSCRIPTEN -D__EMSCRIPTEN_major__=3 -D__EMSCRIPTEN_minor__=1 -D__EMSCRIPTEN_tiny__=8 -fignore-exceptions -fvisibility=default -mllvm -combiner-global-alias-analysis=false -mllvm -enable-emscripten-sjlj -mllvm -disable-lsr -Werror=implicit-function-declaration -Xclang -iwithsysroot/include/SDL --sysroot=/home/linuxbrew/.linuxbrew/Cellar/emscripten/3.1.8/libexec/cache/sysroot -Xclang -iwithsysroot/include/compat -W -Wall -Wextra -Wmissing-prototypes -Wmissing-declarations -Wno-format -Wno-unused-parameter -I../../boards/OSO-SWAT-A1-05 -I../../watch-library/shared/driver/ -I../../watch-library/shared/config/ -I../../watch-library/shared/watch/ -I../../watch-library/simulator/watch/ -I../../watch-library/simulator/hpl/port/ -I../../watch-library/hardware/include/component -I../../watch-library/hardware/hal/include/ -I../../watch-library/hardware/hal/utils/include/ -I../../watch-library/hardware/hpl/slcd/ -I../../watch-library/hardware/hw/ -I../ -I../watch_faces/ -I../watch_faces/clock/ -I../watch_faces/settings/ -I../watch_faces/complication/ -I../watch_faces/sensor/ -I../watch_faces/demo/ -I../lib/TOTP-MCU/ -I../lib/sunriset/ -I../lib/vsop87/ -I../lib/astrolib/ -c ../watch_faces/complication/probability_face.c -o build/probability_face.o' failed (returned 1)
make: *** [../../rules.mk:46: build/probability_face.o] Error 1
emmake: error: 'make' failed (returned 2)
```

You can see the same error happening in the GitHub build-simulator action above. Also as a note I tested with regular `arc4random` and got the same issue, so I don't think it's because I'm using `arc4random_uniform`.


**Notes to self / TODO:**
- [ ] Fix arc4random implicit declaration error
- [ ] Test that `PR` title on watch face sticks around during usage (on emulator)

spencerbyw:
@joeycastillo Pushed a commit to fix the above issue. Basically just added a couple `#if __EMSCRIPTEN__` checks to use regular `rand()` when the code is built for the emulator, and `arc4random` when it's on hardware.

In the (highly likely) scenario there's a better way to resolve the issue, let me know and I'll be happy to make the change :)

joeycastillo:
Interesting! This seems to be [a thing with emscripten](https://github.com/emscripten-core/emscripten/issues/10709), but one of the links in that issue suggests some options for a workaround. I think your `#if __EMSCRIPTEN__` blocks are fine for now, and shortly I'll try to figure out a more universal resolution that'll keep individual watch faces from having to deal with this quirk.

Thanks for the pull request, the watch face looks great!

PR #72, opened 2022-05-19
Adjust probability watch face timeout/LE behavior

spencerbyw:
@joeycastillo I realized yesterday that my probability watch face will timeout and go back to the first watch face after a minute (or whatever's in settings) of inactivity.

I think this is undesirable behavior because my main use case for this is rolling dice during a game, which you'd likely have to do multiple times while playing.

I've changed it to **not** timeout and instead just enter low energy mode after 1 hour (or whatever's in settings).

Question: Is it worth it to enter LE mode if my watch face doesn't really do anything when it's not being used? I'm assuming there are some power savings so I figured why not.

### DIFF
--- a/movement/watch_faces/complication/probability_face.c
+++ b/movement/watch_faces/complication/probability_face.c
@@ -166,8 +166,9 @@ bool probability_face_loop(movement_event_t event, movement_settings_t *settings
             // Dice rolling animation begins on next tick and new roll will be displayed on completion
             movement_request_tick_frequency(PROBABILITY_ANIMATION_TICK_FREQUENCY);
             break;
-        case EVENT_TIMEOUT:
-            movement_move_to_face(0);
+        case EVENT_LOW_ENERGY_UPDATE:
+            watch_display_string("      ", 4);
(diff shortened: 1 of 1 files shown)


tahnok:
I think this works, although you probably don't even need the animation

joeycastillo:
This works! But to @tahnok's point, I also am unsure if the animation is the best way to go here. In other watch faces where there's nothing to display in sleep mode, I've put the word "SLEEP " in the main row instead of the animation. This makes clear that there's nothing updating, and the wearer can press the Alarm button to wake the device. 

Does that seem like a better solution here? Open to feedback; this is one section of the interface guidelines that I haven't fully fleshed out yet.

spencerbyw:
I think that makes sense. I like your idea of when a watch face isn't making any updates in the background, it should show SLEEP, and if it is, show the animation.

That said, I'm too lazy to open my watch again so I'm going to leave the animation on mine 🙃

joeycastillo:
Totally — your watch is yours and it's unique! :)

PR #76, opened 2022-07-03
UV Sensor PCB + Code

tahnok:
Opening this up now so I have a place to push my work in progress in case anyone wants to follow along

@joeycastillo I designed the PCB for the sensor in KiCAD, had to add some new .gitignore stuff and I wasn't sure how to name it to follow the conventions you have... What do you think about a kicad folder?


apps/uv-test/app.c
```diff
@@ -0,0 +1,414 @@
...
+    watch_enable_display();
+
+    watch_enable_external_interrupts();
+    watch_register_interrupt_callback(BTN_MODE, cb_mode_pressed, INTERRUPT_TRIGGER_RISING);
+    watch_register_interrupt_callback(BTN_LIGHT, cb_light_pressed, INTERRUPT_TRIGGER_RISING);
+    watch_register_interrupt_callback(BTN_ALARM, cb_alarm_pressed, INTERRUPT_TRIGGER_RISING);
+
+    watch_enable_leds();
+
+    watch_register_interrupt_callback(A2, cb_si1133_int, INTERRUPT_TRIGGER_FALLING);
```
tahnok, line 230: @joeycastillo can you spot a mistake here somewhere? I can't get the interrupt to fire, even though the pin does go low

joeycastillo: This is a subtle but annoying issue with the ASF external interrupt framework; there's [a warning about it in the docs](https://joeycastillo.github.io/Sensor-Watch/group__buttons.html#gae40d2f73d0412680d83c26b8a0e1d31c). The gist is that pin A2 uses the same interrupt channel as the Alarm button, and the ASF framework only wants one pin associated with each channel. You could change that association [deep in the EIC config file](https://github.com/joeycastillo/Sensor-Watch/blob/main/watch-library/shared/config/hpl_eic_config.h#L726), but it would mean the Alarm button no longer works with the EIC.

My suggestion: instead of using `watch_register_interrupt_callback`, use `watch_register_extwake_callback` ([documented here](https://joeycastillo.github.io/Sensor-Watch/group__deepsleep.html#gacad726423f5f38697c87474ab25244ca)). This has the added benefit that the interrupt on A2 will be able to wake the watch from deep sleep.

tahnok: I guess revision 2 of the UV sensor board should use a different pin :sweat_smile: , thanks for the clarification, I didn't think to read the docs...

joeycastillo: TBH I think pins A2 and A4 are the best option for an external interrupt, since they can wake the watch from deep sleep (the EIC doesn't run in that deep sleep, but the external wake on those two pins does). Unless you need to trigger on both rising and falling edges, the extwake callback on A2 should do everything the interrupt callback would do for you. 

Having said that if you do want to use the EIC for whatever reason, I'd recommend pin A4 since it also gives you the option of waking from deep sleep on an interrupt signal.

PR #77, opened 2022-07-07
Wake Face

joshber:
Basic daily alarm clock face. Seems useful if nothing else in the interest of feature parity with the F-91W’s OEM module, 593.

Also experiments with caret-free UI: One button cycles hours, the other minutes, so there’s no toggling between display and adjust modes and no cycling the caret through the UI.


joeycastillo:
Thanks for opening this pull request; I plan to review this this weekend! At a glance it looks good, my only hesitation is that it intersects with an improvement I want to make (getting the `movement_play_alarm` function to be able to wake the watch from sleep, so a watch face like yours can play a longer alarm tone without having to beep and delay). 

I may want to merge in and then make some changes once I get that API squared away, or do those changes in this PR if that's okay with you?

joeycastillo:
Update: Try merging in the latest changes from main, and replacing the loop in the `EVENT_BACKGROUND_TASK` case with the single line: 

```movement_play_alarm();```

This should play the alarm if the watch is awake when the time matches, and it should also wake it from sleep mode and play the alarm if the watch is in sleep.

Other than that, just two other notes:

* I think it would make sense to illuminate the bell indicator when the alarm is on, instead of using a letter 'o'. The bell icon is more familiar to folks as signaling the status of an alarm.
* I don't think we necessarily need a wakey firmware in the alt_fw directory; in fact I think it makes sense to incorporate the Wakey face into the Focus image, to add an alarm function to that firmware. 

Really great work, thank you for doing this!

joshber:
Love the movement_play_alarm() sound. This should be ready to merge now.

joshber:
🌋 🎡

PR #78, opened 2022-07-14
Replace python with python3 for calling uf2conv

tahnok:
Instead of calling python use python3.

On a default debian 11 install there is no python binary, only python3.
Given that the uf2conv script is written for python 3 I think this is a
safe change to make


joeycastillo:
looks good to me, thanks!

PR #79, opened 2022-07-25
movement: fix movement_request_wake() prototype

jackbondpreston:
This function has `void` args, currently it can throw a compiler warning:

```
../movement.c:225:6: warning: no previous prototype for function 'movement_request_wake' [-Wmissing-prototypes]
void movement_request_wake() {
     ^
../movement.h:291:6: note: this declaration is not a prototype; add 'void' to make it a prototype for a zero-parameter function
void movement_request_wake();
     ^
                           void
```


joeycastillo:
Good catch, thank you!

PR #80, opened 2022-08-02
Movement: add a lil file system with lfs

joeycastillo:
This PR adds a very simple, very small filesystem to Movement. It uses the SAM L22's 8 kilobyte EEPROM emulation area for this purpose, which gives us just shy of 8 kB of usable space. Watch faces can create files, write to them and read from them, which can be useful when you want to store persistent data. This data also persists even when the watch is powered down, making it a more persistent place to stash settings than the BACKUP registers (which reset when the battery is removed). 

Available functions `filesystem_init`, `filesystem_get_free_space`, `filesystem_file_exists`, `filesystem_rm`, `filesystem_get_file_size`, `filesystem_read_file`, and `filesystem_write_file` are documented in `filesystem.h` in the Movement folder. There is room for improvement here; right now these functions require that you have your whole file loaded in memory, and don't allow for things like append operations (which lfs supports). Future improvements would be welcome, but after leaving this branch open for months, I figure this is good enough to get started.

You can't mount this filesystem as a USB drive or anything, but the Movement firmware supports very basic file management at the USB serial terminal. Preliminary documentation: 

* `ls` prints a listing of files at the root. At this time, the USB serial file manager only works at the root of the filesystem.
* `cat file` prints out a single file. I know, I know.
* `rm file` removes a file.
* `echo text > file` writes a single line of text to a file.
* `df` shows the free space on the filesystem.


tahnok:
Fantastic timing, I've got some sensor data I need to log to calibrate the
UV sensor

PR #84, opened 2022-08-29
add support for multiple TOTP keys

jackbondpreston:
seems to work fine on the simulator and hardware

### DIFF
--- a/movement/watch_faces/complication/totp_face.c
+++ b/movement/watch_faces/complication/totp_face.c
@@ -1,7 +1,3 @@
-/**
- * TODO:
- * - Support for multiple codes
- */
 #include <stdlib.h>
 #include <string.h>
 #include "totp_face.h"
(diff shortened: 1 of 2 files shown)


movement/watch_faces/complication/totp_face.c
```diff
@@ -64,8 +74,18 @@ bool totp_face_loop(movement_event_t event, movement_settings_t *settings, void
...
         case EVENT_ALARM_BUTTON_UP:
+            if (index + 1 < num_keys) {
+                totp_state->current_key_offset += key_sizes[index];
+                totp_state->current_index++;
+            } else {
+                // wrap around to first key
+                totp_state->current_key_offset = 0;
+                totp_state->current_index = 0;
+            }
+            TOTP(keys + totp_state->current_key_offset, key_sizes[totp_state->current_index], timesteps[totp_state->current_index]);
```
tahnok, line 86: IMO, it's a bit clearer to use a subscript here rather than adding an offset

tahnok (approved):
Nice, thanks for doing this @jackbondpreston !

I agree, using LFS would be a better place for keys, you might even be able to have a command line interface to set up the secret

PR #86, opened 2022-10-06
Fix changed files detection in Make

michaelgruenewald:
Detecting files changes doesn't work for me in the movement project. Whenever I change a source file, I have to run `make clean` and `make` to compile my changes.

AFAICT this is because of a broken expansion of the `DEPFILES` variable. If I add `$(warning DEPFILES=$(DEPFILES))` to the end of _rules.mk_ and run `make` in _movement/make_, I get something like

    ../../rules.mk:67: DEPFILES=./build/../../tinyusb/src/tusb.d ... ./build/../watch_faces/demo/hello_there_face.d ...

It somehow assumes that the .o.d. files have the same directory structure as the source files, but they are actually flat in the `$(BUILD)` directory. Consequently, adding `$(warning DEPFILES=$(wildcard $(DEPFILES)))` to _rules.mk_ just outputs

    ../../rules.mk:67: DEPFILES=

which just means no dependency files are actually included.

This change makes it derive the dependency file names from the object file names, which seems to work fine for me. Change detection works now, building all the projects from clean and dirty states also works for me.

### DIFF
--- a/make.mk
+++ b/make.mk
@@ -159,6 +159,7 @@ else
 
 CFLAGS += -W -Wall -Wextra -Wmissing-prototypes -Wmissing-declarations
 CFLAGS += -Wno-format -Wno-unused-parameter
+CFLAGS += -MD -MP -MT $(BUILD)/$(*F).o -MF $(BUILD)/$(@F).d
 
 INCLUDES += \
   -I$(TOP)/boards/$(BOARD) \
--- a/rules.mk
+++ b/rules.mk
@@ -1,3 +1,5 @@
+.SUFFIXES:
+
 CFLAGS += $(INCLUDES) $(DEFINES)
 
 OBJS = $(addprefix $(BUILD)/, $(notdir %/$(subst .c,.o, $(SRCS))))
--- a/rules.mk
+++ b/rules.mk
@@ -60,6 +60,6 @@ clean:
 analyze:
 	@$(COBRA) basic $(INCLUDES) $(DEFINES) $(SRCS)
 
-DEPFILES := $(SRCS:%.c=$(BUILD)/%.d)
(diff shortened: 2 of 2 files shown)


tahnok:
Nice! I haven't had a chance to test this yet, but it's been bugging me as well. I think c40d75b8765a2972bb6a80a7c6f56865b52063e4 introduced it and it has something to do with order-only dependencies (which I don't fully understand)

michaelgruenewald:
I also enabled dependency file generation for the emscripten HTML build where it wasn't enabled before.

joeycastillo:
First off thank you so much for doing this! I had noticed the issue in simulator builds from the outset, and in builds for the hardware recently, so your diagnosis of when it started to happen tracks for me. 

One issue: in my testing (MacOS Monterrey), `make` seems to run fine the first time around, but then hang indefinitely on subsequent builds. Running `make -d` to output debug info seems to suggest it's caught in an infinite loop. [Here's one megabyte of the log from an attempted build](https://github.com/joeycastillo/Sensor-Watch/files/9740136/build.log); the infinite loop seems to start around line 6,300.

michaelgruenewald:
Hrmm, interesting. I'm not a Make expert by any means, but comparing your log with my good log (from Linux), it looks like your make somehow thinks that it can make `.o.d.` files from `.o.d.o` files, which doesn't make any sense, but it still tries.

I added a `.SUFFIXES:` rule on top which is supposed to disable all built-in rules (which we don't need anyway). I wonder whether that fixes it.

joeycastillo:
This does indeed seem to address it! Thanks again for fixing this issue; it's definitely going to help folks as they build their own firmware images :)

PR #93, opened 2022-10-12
Make starter project work in simulator

wryun:
Use the higher abstraction level sleeping, since the simulator (emulator?) doesn't understand the lower level HAL.

I'm assuming this is a reasonable thing to do...

### DIFF
--- a/apps/starter-project/app.c
+++ b/apps/starter-project/app.c
@@ -1,6 +1,7 @@
 #include <stdio.h>
 #include <string.h>
 #include "watch.h"
+#include "watch_main_loop.h"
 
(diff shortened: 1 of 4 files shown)


wryun:
Well, I'm a little confused here about why it doesn't work in CI's docker image and does for me locally (and vice-versa, apparently).

joeycastillo:
No worries! Your change fixed the simulator build; we just needed to #ifdef out that include for builds on hardware. I moved the conditional include to `watch.h`, and added a `delay_ms` function for the simulator. 

Thanks so much for catching this!

PR #94, opened 2022-10-12
Long press mode button to return to first watch face + fix lag when watch face changes

wryun:
This is the simple version of #89 

I'm in two minds about this. It seems like a neat hack, but I think a better solution would be to have the loop() functions return whether they want to capture/bubble the event (cf Javascript), in which case we could handle _all_ the standard functions here in a consistent way without relying on the faces.

i.e. EVENT_MODE_BUTTON_UP => movement_move_to_next_face(), EVENT_MODE_LONG_PRESS/EVENT_TIMEOUT => movement_move_to_face(0),  EVENT_LIGHT_BUTTON_DOWN => movement_illuminate_led()

The return from 'loop' would be something like:

```
struct {
  bool can_sleep;
  bool event_swallowed;
}
```

@josecastillo - how would you feel about this refactor?

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -417,7 +417,6 @@ bool app_loop(void) {
             && movement_state.current_watch_face > 0 
             && movement_state.current_watch_face == movement_state.next_watch_face) {
             movement_move_to_face(0);
-            movement_state.watch_face_changed = true;
         }
         event.event_type = EVENT_NONE;
     }
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -407,9 +407,14 @@ bool app_loop(void) {
     if (event.event_type) {
         event.subsecond = movement_state.subsecond;
         can_sleep = watch_faces[movement_state.current_watch_face].loop(event, &movement_state.settings, watch_face_contexts[movement_state.current_watch_face]);
-        // escape hatch: a watch face may not resign on EVENT_MODE_BUTTON_DOWN. In that case, a long press of MODE should let them out.
-        if (event.event_type == EVENT_MODE_LONG_PRESS) {
-            movement_move_to_next_face();
+
+        // Long-pressing MODE brings one back to the first face, provided that the watch face hasn't decided to send them elsewhere
+        // (and we're not currently on the first face).
+        // Note that it's the face's responsibility to provide some way to get to the next face, so if EVENT_MODE_BUTTON_* is
+        // used for face functionality EVENT_MODE_LONG_PRESS should probably be handled and next_face() triggered in the face
+        // (which would effectively disable the normal 'long press to face 0' behaviour).
+        if (event.event_type == EVENT_MODE_LONG_PRESS && movement_state.current_watch_face > 0 && movement_state.current_watch_face == movement_state.next_watch_face) {
+            movement_move_to_face(0);
             can_sleep = false;
         }
         event.event_type = EVENT_NONE;
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -413,9 +413,11 @@ bool app_loop(void) {
         // Note that it's the face's responsibility to provide some way to get to the next face, so if EVENT_MODE_BUTTON_* is
         // used for face functionality EVENT_MODE_LONG_PRESS should probably be handled and next_face() triggered in the face
         // (which would effectively disable the normal 'long press to face 0' behaviour).
-        if (event.event_type == EVENT_MODE_LONG_PRESS && movement_state.current_watch_face > 0 && movement_state.current_watch_face == movement_state.next_watch_face) {
+        if (event.event_type == EVENT_MODE_LONG_PRESS 
+            && movement_state.current_watch_face > 0 
+            && movement_state.current_watch_face == movement_state.next_watch_face) {
             movement_move_to_face(0);
-            can_sleep = false;
+            movement_state.watch_face_changed = true;
         }
         event.event_type = EVENT_NONE;
     }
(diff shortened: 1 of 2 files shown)


wryun:
The other thing I'm not clear about is what exactly is going on with can_sleep here. It does make sense to me not to sleep the watch when switching faces, but if I look at the other faces (that do switches) can_sleep is usually set to true. I have noticed some slow behaviour on switching on occasion, which this might be the cause of.

Should the logic be something like:

```
if (current_face != next_face) {
  can_sleep = false; 
}
```
?

Or is there some other reason for can_sleep?

joeycastillo:
Thanks for this! I'm going to put together a build with this feature and wear it for a bit. 

> The other thing I'm not clear about is what exactly is going on with can_sleep here.

So, `can_sleep` is the core of the low power capabilities of Sensor Watch. If the watch face returns `true` — which it should 99% or more of the time — the watch can enter standby mode until woken up by an interrupt. In wake mode that's the 1 Hz tick or a button press; in low energy mode, that's the one minute alarm. The only times a watch face returns `false` is if it's doing something like dimming an LED that requires the main clock to keep running; in this case, the loop function gets called repeatedly as long as `can_sleep` comes back true. 

Usually we don't have to touch this variable; once we're out of the watch face loop, we know the watch face has set the display to show what it wants to show, and we don't need to ask it again until the next tick or button press. In this case, though, the watch face has changed but the display is still showing whatever the previous watch face had displayed. So we need to set `can_sleep` to `false`, which causes the loop to execute again immediately. This calls the new watch face's activate and loop functions, after which the new face (hopefully) returns `true` so we can go right back to sleep.

wryun:
@joeycastillo but when I look at the code for a face is does movement_move_to_next_face (or indeed movement_move_to_face(0)) and sets can_sleep to true. Why is this any different to what we're doing here? The watch face hasn't been updated yet! Surely we should set can_sleep=false after any movement, so we immediately display the next face?

wryun:
(relatedly, I've now noticed the watch_face_changed variable, which looks pretty similar to my check here. Probably I should replace my check with this, or we should remove the watch_face_changed variable as unnecessary (?) )

joeycastillo:
> when I look at the code for a face it does movement_move_to_next_face (or indeed movement_move_to_face(0)) and sets can_sleep to true

You're quite right. It looks like I've been doing this wrong in a ton of watch faces, which quite likely to be the source of some of the lagginess beyond the simple clock and world clock faces. Watch faces should return `false` after calling `movement_move_to_next_face`, but it's a subtle enough footgun that I built the foot and the gun and still managed to do some damage.

If you don't mind, I'll push a change to this branch that uses `watch_face_changed` to do this check at the end.

wryun:
:+1: 

I think that this is the sort of thing that would be less likely to happen with the refactor I proposed (hint, hint) ;)

joeycastillo:
I would really like to avoid significant refactors at this juncture; I want to avoid major changes that might lead to unexpected behavior, or worse, a hang or a crash. Since we can't use the watchdog, the wearer has no way to recover from a hang without disassembling the watch; this makes me very risk-averse when it comes to the core loop. Fixing the lagginess feels necessary, and the return home gesture feels low-risk, but still I'm going to want to test this change for at least a couple of weeks before I feel comfortable landing it. A refactor like what you're describing is something I'd want to spend a lot more time on.

wryun:
Fair call - thanks for explaining.

joeycastillo:
Made the change, and made the check at the end of the loop a bit less inscrutable. Initial tests are promising; let me know how you find it!

movement/movement.c
```diff
@@ -407,10 +407,17 @@ bool app_loop(void) {
...
-            can_sleep = false;
+
+        // Long-pressing MODE brings one back to the first face, provided that the watch face hasn't decided to send them elsewhere
+        // (and we're not currently on the first face).
+        // Note that it's the face's responsibility to provide some way to get to the next face, so if EVENT_MODE_BUTTON_* is
+        // used for face functionality EVENT_MODE_LONG_PRESS should probably be handled and next_face() triggered in the face
+        // (which would effectively disable the normal 'long press to face 0' behaviour).
+        if (event.event_type == EVENT_MODE_LONG_PRESS 
+            && movement_state.current_watch_face > 0 
+            && movement_state.current_watch_face == movement_state.next_watch_face) {
```
wryun, line 418: Possibly this condition should change to `!movement_state.watch_face_changed` (sorry, didn't realise that this existed initially)

movement/movement.c
```diff
@@ -476,7 +483,13 @@ bool app_loop(void) {
 
     event.subsecond = 0;
 
-    return can_sleep && (movement_state.light_ticks == -1) && !movement_state.is_buzzing;
+    // if the watch face changed, we can't sleep because we need to update the display.
```
wryun, line 486: Nice, this whole section is now much clearer. For 'consistency', I guess we could remove the 'return false' after move_to_next_face in simple_clock_face, world_clock_face, and preferences_face.

movement/movement.c
```diff
@@ -407,10 +407,17 @@ bool app_loop(void) {
...
+        // Long-pressing MODE brings one back to the first face, provided that the watch face hasn't decided to send them elsewhere
+        // (and we're not currently on the first face).
+        // Note that it's the face's responsibility to provide some way to get to the next face, so if EVENT_MODE_BUTTON_* is
+        // used for face functionality EVENT_MODE_LONG_PRESS should probably be handled and next_face() triggered in the face
+        // (which would effectively disable the normal 'long press to face 0' behaviour).
+        if (event.event_type == EVENT_MODE_LONG_PRESS 
+            && movement_state.current_watch_face > 0 
+            && movement_state.current_watch_face == movement_state.next_watch_face) {
+            movement_move_to_face(0);
+            movement_state.watch_face_changed = true;
```
wryun, line 420: This look unnecessary (move_to_face sets this). I'd still prefer to remove this variable entirely and rely on the comparison everywhere, but I guess this falls into the bucket of overly drastic changes :)

joeycastillo: Good call, can remove that redundant line.

movement/movement.c
joeycastillo: I don't think it matters tbh; I added [documentation](https://github.com/joeycastillo/Sensor-Watch/pull/94/commits/f89cecc56a974b914b54a190157bd77d1e7c9be9) noting that the return value is moot once you call `movement_move_to_next_face`.

joeycastillo:
I've been testing this for a few days now (alongside at least a couple of folks in the Discord) and haven't noticed or heard of any issues. Feeling good about this; merging.

PR #95, opened 2022-10-13
Make the TOTP face use the filesystem for secret storage

wryun:
Messed around with using LFS more nicely, then decided to stop getting distracted since the memory usage from reading the entire file in was acceptable on startup.

Only limited testing on simulator so far, pending feedback on approach. Don't merge as is.

Why this is good:

- don't have to bake TOTP keys into code
- understands normal totp format
- understands files exported from Aegis

Why this is bad:

- should we be encrypting? Probably no point (unless one can't easily extract the firmware? i.e. that it's a lot easier to pull this file that the code?)
- 'parsing' code is a bit sketchy
- not clear whether we want the watch to be doing so much work (could have a script to 'upload' the file into a form that's closer or identical to the in-memory representation)
- with base32.c, there's a bunch more code (could easily cut the encode function from the base32 library, I guess)


movement/watch_faces/complication/totp_face.c
```diff
@@ -1,44 +1,180 @@
...
+static bool totp_face_read_param(struct totp_record *totp_record, char *param, char *value) {
+    if (!strcmp(param, "issuer")) {
+        if (value[0] == '\0' || value[1] == '\0') {
+            printf("TOTP issuer must be >= 2 chars, got '%s'", value);
+            return false;
+        }
+        totp_record->label[0] = value[0];
+        totp_record->label[1] = value[1];
+    } else if (!strcmp(param, "secret")) {
+        totp_record->secret = malloc(ceil(strlen(value) / 5.0) * 8);
```
wryun, line 63: TODO: should have used BASE32_LEN

movement/watch_faces/complication/totp_face.c
```diff
@@ -1,44 +1,180 @@
...
-    totp_state->timestamp = watch_utility_date_time_to_unix_time(watch_rtc_get_date_time(), movement_timezone_offsets[settings->bit.time_zone] * 60);
-    totp_state->current_code = getCodeFromTimestamp(totp_state->timestamp);
+
+    if (num_totp_records > 0) {
+        totp_state->current_index = 0;
+        totp_state->steps = 0;  // Force getting a new code in the loop.
+        totp_state->timestamp = watch_utility_date_time_to_unix_time(watch_rtc_get_date_time(), movement_timezone_offsets[settings->bit.time_zone] * 60);
+        TOTP(totp_records[0].secret, totp_records[0].secret_size, totp_records[0].period);
+    } else {
+        watch_display_string("NA00000000", 0);
```
wryun, line 176: Suggestions welcome for a what a better string is here.

movement/watch_faces/complication/totp_face.c
```diff
@@ -75,15 +219,11 @@ bool totp_face_loop(movement_event_t event, movement_settings_t *settings, void
...
-            } else {
-                // wrap around to first key
-                totp_state->current_key_offset = 0;
-                totp_state->current_index = 0;
+            if (num_totp_records == 0) {
+                break;
             }
-            TOTP(keys + totp_state->current_key_offset, key_sizes[totp_state->current_index], timesteps[totp_state->current_index]);
+            totp_state->current_index = (index + 1) % num_totp_records;
+            TOTP(totp_records[totp_state->current_index].secret, totp_records[totp_state->current_index].secret_size, totp_records[totp_state->current_index].period);
```
wryun, line 226: TODO: line looks a bit long. Probably should just set 'index' as above. Hmm, and feels like I should reset steps to force getCodeFromTimestamp to happen again.

joeycastillo:
Thanks for this! It's a good start. I'm going to get in there and make some changes, though. 

My main issues center around the extra malloc's and free's that are happening here. My vision for Movement is that watch faces generally shouldn't call malloc except the once, in setup, to claim space for their watch face state. At this time all watch faces have one call to malloc, and no call to free; this effectively means all memory is statically allocated at boot and sticks around in exactly the same place until the watch battery dies. 

I think there is a refactor that will allow us to do that here in the TOTP face, but I'd prefer to hack on it myself (which will also give me a chance to review the PR in depth).

wryun:
Thanks! I'll be interested to see how you get around not using malloc. The only thing I can really think of is to read the file into a static buffer and then just point to the secrets in that buffer, but that'll end up wasting quite a bit of space relatively (vs malloc/freeing just in the init)...

wryun:
When I finally tried to flash this to my watch, there were a couple of issues:
- the serial connection was a bit problematic because of the way 'read' is working. Using picocom gave me character at a time into 'read' (i.e. couldn't execute any commands, since there's essentially no line buffering); the Arduino IDE's serial monitor allowed me to send almost a line in a quick chunk, but over 50 bytes got chopped off in the middle. i.e. it's pretty hard to get the secrets across.
- when you get the secrets across, you get a different sequence to the emulator. I've tried to be pretty careful about making sure the time was set properly, yet I persistently get the wrong numbers. I was getting the _correct_ numbers before this PR though when I hardcoded my secrets. Unfortunately, this is hard to debug, because 'cat' doesn't work either across the serial connection for larger files (you just get the first 50 or so bytes).

joeycastillo:
OKAY! I've made the changes we discussed: static allocation of all the secrets (it uses up a little over a kilobyte of RAM, which could worry me but not right now) and it reads the file line by line into a fixed-length buffer (max line length is 255). I also renamed the watch face to `totp_face_lfs`; it occurs to me that some folks will still want to do it the old way, compiling their secrets into the firmware, so we should keep both watch faces available. 

Feel free to play around with this and let me know if you see any issues; when you're ready, mark as ready for review and I can merge it in :)

movement/filesystem.c
```diff
@@ -166,6 +166,28 @@ bool filesystem_read_file(char *filename, char *buf, int32_t length) {
...
+        err = lfs_file_seek(&lfs, &file, offset, LFS_SEEK_SET);
+        if (err < 0) return false;
+        err = lfs_file_read(&lfs, &file, buf, min(length, file_size - offset));
+        if (err < 0) return false;
+        for(int i = 0; i < length; i++) {
+            if (buf[i] == '\n') {
+                buf[i] = 0;
+                break;
+            }
+        }
```
tahnok, line 184: I feel like there's a subtle bug here where you don't get null terminated string if you read length and don't hit a new line or if your file is less than length and it doesn't end in a newline

movement/filesystem.c
```diff
@@ -166,6 +166,28 @@ bool filesystem_read_file(char *filename, char *buf, int32_t length) {
...
+    memset(buf, 0, length);
+    int32_t file_size = filesystem_get_file_size(filename);
+    if (file_size > 0) {
+        int err = lfs_file_open(&lfs, &file, filename, LFS_O_RDONLY);
+        if (err < 0) return false;
+        err = lfs_file_seek(&lfs, &file, offset, LFS_SEEK_SET);
+        if (err < 0) return false;
+        err = lfs_file_read(&lfs, &file, buf, min(length, file_size - offset));
+        if (err < 0) return false;
+        for(int i = 0; i < length; i++) {
```
wryun, line 179: Or:

```
char *end_of_line = strchr(buf, '\n');
if (end_of_line)
  *end_of_line = '\0';
```

I think. But if the function signature is changed to return the offset, then I guess there's more pay-off in doing our own loop.

movement/watch_faces/complication/totp_face_lfs.c
```diff
@@ -0,0 +1,231 @@
...
+static bool totp_face_lfs_read_param(struct totp_record *totp_record, char *param, char *value) {
+    if (!strcmp(param, "issuer")) {
+        if (value[0] == '\0' || value[1] == '\0') {
+            printf("TOTP issuer must be >= 2 chars, got '%s'\n", value);
+            return false;
+        }
+        totp_record->label[0] = value[0];
+        totp_record->label[1] = value[1];
+    } else if (!strcmp(param, "secret")) {
+        totp_record->secret_size = base32_decode((unsigned char *)value, totp_record->secret);
```
wryun, line 63: I feel like checking BASE32_LEN is the responsible thing to do here... (i.e. make sure it's <= 48, now that the max secret size is fixed)

movement/watch_faces/complication/totp_face_lfs.c
```diff
@@ -0,0 +1,231 @@
...
+        }
+    }
+}
+
+void totp_face_lfs_setup(movement_settings_t *settings, uint8_t watch_face_index, void ** context_ptr) {
+    (void) settings;
+    (void) watch_face_index;
+    if (*context_ptr == NULL) {
+        *context_ptr = malloc(sizeof(totp_lfs_state_t));
+        totp_face_lfs_read_file(TOTP_FILE);
```
wryun, line 152: Theoretically, to support multiple TOTP faces, I guess this should be in a separate 'if' to check if our static data has been initialised. But in practice it would only make sense to have multiple faces if we had different TOTP files, I think.

movement/watch_faces/complication/totp_face_lfs.c
```diff
@@ -0,0 +1,231 @@
...
+            break;
+        case EVENT_TIMEOUT:
+            movement_move_to_face(0);
+            break;
+        case EVENT_ALARM_BUTTON_UP:
+            if (num_totp_records == 0) {
+                break;
+            }
+            totp_state->current_index = (index + 1) % num_totp_records;
+            TOTP(totp_records[totp_state->current_index].secret, totp_records[totp_state->current_index].secret_size, totp_records[totp_state->current_index].period);
```
wryun, line 217: Having used the TOTP face and understood more about what's going on, it would be good to do the display here. i.e. probably pull the ACTIVATE code into a display function, and call is from TICK/ACTIVATE/ALARM_BUTTON_UP.

wryun (commented):
Looks good.

Some thoughts; haven't run the code yet. Happy to make suggested changes if they seem ok.

wryun:
Re keeping the old TOTP face, is there a reason to prefer that approach aside from a bit of 'security through obscurity' for the secrets?

movement/filesystem.c
joeycastillo: Good call; it works as long as you have a buffer one byte longer than the maximum length you're requesting, but it's probably safer to just max out at `length - 1`.

movement/filesystem.c
```diff
@@ -166,6 +166,28 @@ bool filesystem_read_file(char *filename, char *buf, int32_t length) {
     return false;
 }
 
+bool filesystem_read_line(char *filename, char *buf, int32_t offset, int32_t length) {
```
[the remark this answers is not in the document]
joeycastillo: I understand the desire to return the number of bytes read, but not sure I understand "take offset as a point and do the update itself"? Take offset as a line offset rather than a byte offset?

movement/filesystem.c
```diff
@@ -166,6 +166,28 @@ bool filesystem_read_file(char *filename, char *buf, int32_t length) {
...
+    memset(buf, 0, length);
+    int32_t file_size = filesystem_get_file_size(filename);
+    if (file_size > 0) {
+        int err = lfs_file_open(&lfs, &file, filename, LFS_O_RDONLY);
+        if (err < 0) return false;
+        err = lfs_file_seek(&lfs, &file, offset, LFS_SEEK_SET);
+        if (err < 0) return false;
+        err = lfs_file_read(&lfs, &file, buf, min(length, file_size - offset));
```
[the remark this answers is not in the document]
joeycastillo: Yep, this seems like the move :)

movement/watch_faces/complication/totp_face_lfs.c
joeycastillo: This is also a good call, I can make this change.

movement/watch_faces/complication/totp_face_lfs.c
joeycastillo: yea I don't foresee multiple TOTP faces really making sense; since the file is hard coded, I think it's reasonable to say "have a maximum of one TOTP face in the list"

joeycastillo:
> Re keeping the old TOTP face, is there a reason to prefer that approach aside from a bit of 'security through obscurity' for the secrets?

The main reason is that there are already [blog posts out there](https://blog.singleton.io/posts/2022-10-17-otp-on-wrist/) documenting the current workflow, and I am hesitant to break things for folks who are already using the existing functionality as provided at launch. I think both faces present valid options for folks who want to use this function, and there's no harm in including both; watch faces don't take up any flash or RAM unless included in the list;

movement/watch_faces/complication/totp_face_lfs.c
joeycastillo: Def agree; this works, but lags by up to one second while it waits for the next tick.

movement/filesystem.c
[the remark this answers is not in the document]
wryun: Sorry, point = pointer (those letters would have been useful. As in, pass the parameter:

```
int32_t *offset;
```

Then, we can be responsibility for updating the offset, and the caller just has to provide it on subsequent calls without modification.

cf strtok_r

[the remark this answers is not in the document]
joeycastillo: Uncertain about this, mainly because it means you have to read a file line by line every time. Let's say you have a file with lines of data (i.e. arrival times for each subway station on a line). Your watch face could stash the relevant station as an offset into the file, and know that it needs to read a line at offset X to get all the arrivals for Main Street. 

If I understand the suggestion correctly, the offset pointer would only be valid in the context of parsing lines sequentially, and then only once; whereas stating that you want to read a line at an arbitrary offset leverages the ability to seek to any set point and read bytes. This feels, to me, like a more generally useful form for the function to take.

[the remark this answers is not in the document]
wryun: It would be the same value, just stored/used in the location you provided, so you could do:

```
int offset = 5;
filesystem_read_line(f, buf, &offset, length)
```

Then `*offset` would have the new offset.

i.e. the normal case is 'caller doesn't care what the offset is', but if the caller does care they still have access.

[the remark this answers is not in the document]
joeycastillo: I see, you’re suggesting to make offset an in/out parameter. That makes sense.

wryun:
I've made most of the changes discussed except the filesystem API one. Haven't resolved in case you want to go through the comments.

wryun:
Also, `totp_face_lfs` doesn't follow the normal naming convention (`totp_lfs_face`?). Not sure if that's intentional or not.

movement/watch_faces/complication/totp_face.c
joeycastillo: I changed this to read "No 2F codes" across both lines.

movement/filesystem.c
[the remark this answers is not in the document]
joeycastillo: Implemented this; *offset now gets set to the offset of the next line automatically, removing the need for the strlen call. Thanks for the suggestion!

movement/filesystem.c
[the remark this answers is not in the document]
joeycastillo: Instead of reading to length - 1, I clarified the documentation to state that you should provide a buffer of length + 1, and that the last character will be a null terminator; this is how we were currently using the function anyway, and it feels more obvious than reading one character less than we were asked for.

joeycastillo:
Okay! I made the changes we discussed; take a look and if you think I haven't mangled it up too badly, I can merge it in :)

Upon further reflection, I think you were correct to use malloc instead of a fixed maximum secret size, so I restored that logic. I had gotten spooked because the call was happening in `activate` instead of `setup`, but since it checks against the static number of codes, it is effectively only called once. Just to be sure, I used a pair of #ifdefs so that the file is normally read in `setup`, and only read in `activate` on the simulator; this ensures that the mallocs in that function will only get called once at boot. 

> Also, `totp_face_lfs` doesn't follow the normal naming convention (`totp_lfs_face`?). Not sure if that's intentional or not.

This doesn't bother me; it's a variant of a face, I could see e.g. `temperature_face_thermistor` and `temperature_face_i2c`; as long as the meaning is clear and nothing in the header steps on another watch face's namespace, it should be fine.

movement/watch_faces/complication/totp_face_lfs.c
```diff
@@ -0,0 +1,253 @@
...
+            return false;
+        }
+        totp_record->label[0] = value[0];
+        totp_record->label[1] = value[1];
+    } else if (!strcmp(param, "secret")) {
+        if (UNBASE32_LEN(strlen(value)) > MAX_TOTP_SECRET_SIZE) {
+            printf("TOTP secret too long: %s\n", value);
+            return false;
+        }
+        totp_record->secret = malloc(BASE32_LEN(strlen(value)));
```
wryun, line 68: Sorry, my mistake earlier - this should be UNBASE32_LEN (as above).

wryun (commented):
Thanks - LGTM aside from my mistake with BASE32_LEN vs UNBASE32_LEN.

bademux:
Hello, is it ready for testing? 
Will get grateful for quick manual how to configure it with FS

wryun:
@bademux unfortunately, due to issues with the serial interface to the Sensor Watch, your best bet is to test it in the simulator (I do have it on my watch, but it wasn't fun getting to that point; the easiest thing to do is probably hack a watch face into the firmware to write the necessary files, then remove that face!). There are instructions at the top of the the face for how to use it:

https://github.com/joeycastillo/Sensor-Watch/blob/9ad4184fb40b3f07eaba0f26a56b27af63eba852/movement/watch_faces/complication/totp_face_lfs.c#L22

wryun:
(I believe it's basically ready to merge, aside from the issue with BASE32_LEN - which just causes it to use a bit more space than it should)

joeycastillo (approved):
Right on, I think we're there! Thanks for all your hard work on this one @wryun :)

bademux:
@wryun Is the serial console something that can be temporary hackfixed from client side (some app that will send totp uri in a special way),
or it have to be fixed in the firmware @joeycastillo.
Thanks for the great software and hardware!

joeycastillo:
It’s something that will need a firmware fix, @wryun if you can characterize the issue with some specifics, can you open an issue? I know it has to do with my USB code, and might require someone with a better knowledge of tinyUSB to dig into it (or I’ll take a look eventually)

wryun:
Thanks @joeycastillo ! I've raised https://github.com/joeycastillo/Sensor-Watch/issues/117

As mentioned there, there's a relatively straightforward issue in movement.c, but I suspect there are also deeper problems with how TinyUSB works. But I may be misremembering; I think I got dissuaded from doing the simple fix after hitting other issues.

PR #96, opened 2022-10-14
Add an Alarm face to movement

TheOnePerson:
Implements 16 alarm slots for the sensor watch.

**How to use the alarm face:**
- In normal mode, the alarm button cycles through all 16 alarms. 
- Long pressing the alarm button in normal mode toggles the corresponding alarm on or off.
- Pressing the light button enters setting mode and cycles through the settings of each alarm.
- In setting mode an alarm slot is selected by pressing the alarm button when the slot number 
  in the upper right corner is blinking.
- For each alarm slot, you can select the day. These are the day modes:
	- ED = the alarm rings every day
	- 1t = the alarm fires only one time and is erased afterwards
	- MF = the alarm fires Mondays to Fridays
	- WN = the alarm fires on weekends (Sa/Su)
	- MO to SU = the alarm fires only on the given day of week
- You can fast cycle through hour or minute setting via long press of the alarm button.
- You can select the tone in which the alarm is played. (Three pitch levels available)
- You can select how many "beep rounds" are played for each alarm (1 to 9 beep rounds plus extra short [o] and extra long [L]).
- The simple watch face indicates any alarm set by showing the bell indicator.

**Technical Stuff:**

There are changes in `movement.c`, `movement.h`, and `simple_clock_face.c` to make customizable alarm lengths and pitch levels possible, and also to be able to show an alarm indicator on the simple watch face.
- `movement_settings_t` has an additional flag `alarm_enabled`.
- `movement_state_t` has an additional field `alarm_note`.
- I added the function `movement_play_alarm_beeps(uint8_t rounds, BuzzerNote alarm_note)` to `movement.c` to play alarms of customizable length and tone.
- `simple_clock_face.c` now indicates if `alarm_enabled` is set with the bell indicator.

Edit: Bump number of alarms to 16 and mention extra short and extra long alarms.

### DIFF
--- a/movement/movement.h
+++ b/movement/movement.h
@@ -61,7 +61,8 @@ typedef union {
         // altimeter to display feet or meters as easily as it tells a thermometer to display degrees in F or C.
         bool clock_mode_24h : 1;            // indicates whether clock should use 12 or 24 hour mode.
         bool use_imperial_units : 1;        // indicates whether to use metric units (the default) or imperial.
-        uint8_t reserved : 7;               // room for more preferences if needed.
+        bool alarm_enabled : 1;             // indicates wheter there is at least one alarm enabled.
+        uint8_t reserved : 6;               // room for more preferences if needed.
     } bit;
     uint32_t reg;
 } movement_settings_t;
@@ -248,6 +249,7 @@ typedef struct {
     // alarm stuff
     int16_t alarm_ticks;
     bool is_buzzing;
+    BuzzerNote alarm_note;
 
     // button tracking for long press
     uint8_t light_down_timestamp;
@@ -292,6 +294,7 @@ void movement_request_wake(void);
 
 void movement_play_signal(void);
 void movement_play_alarm(void);
+void movement_play_alarm_beeps(uint8_t rounds, BuzzerNote alarm_note);
 
 uint8_t movement_claim_backup_register(void);
 
--- a/movement/movement_config.h
+++ b/movement/movement_config.h
@@ -30,7 +30,6 @@
 const watch_face_t watch_faces[] = {
     simple_clock_face,
     world_clock_face,
-    alarm_face,
     sunrise_sunset_face,
     moon_phase_face,
     thermistor_readout_face,
@@ -39,5 +38,12 @@ const watch_face_t watch_faces[] = {
 };
 
 #define MOVEMENT_NUM_FACES (sizeof(watch_faces) / sizeof(watch_face_t))
+
+/* Determines what face to go to from the first face if you've already set 
+ * a mode long press to go to the first face in preferences, and
+ * excludes these faces from the normal rotation.
+ * Usually it makes sense to set this to the preferences face.
+ */
+#define MOVEMENT_SECONDARY_FACE_INDEX 0 // or (MOVEMENT_NUM_FACES - 2)
 
 #endif // MOVEMENT_CONFIG_H_
--- a/movement/movement_config.h
+++ b/movement/movement_config.h
@@ -30,6 +30,7 @@
 const watch_face_t watch_faces[] = {
     simple_clock_face,
     world_clock_face,
+    alarm_face,
     sunrise_sunset_face,
     moon_phase_face,
     thermistor_readout_face,
--- a/movement/watch_faces/complication/alarm_face.c
+++ b/movement/watch_faces/complication/alarm_face.c
@@ -89,9 +89,10 @@ static void _alarm_face_draw(movement_settings_t *settings, alarm_state_t *state
     //handle am/pm for hour display
     uint8_t h = state->alarm[state->alarm_idx].hour;
     if (!settings->bit.clock_mode_24h) {
-        if (h > 12) {
+        if (h >= 12) {
             watch_set_indicator(WATCH_INDICATOR_PM);
-            h -= 12;
+            h = h % 12;
+            h += h ? 0 : 12;
         } else {
             watch_clear_indicator(WATCH_INDICATOR_PM);
         }
@@ -102,7 +103,7 @@ static void _alarm_face_draw(movement_settings_t *settings, alarm_state_t *state
         h,
         state->alarm[state->alarm_idx].minute);
     // blink items if in settings mode
-    if (state->is_setting && subsecond % 2 && state->setting_state < alarm_setting_idx_pitch && state->alarm_quick_ticks == -1) {
+    if (state->is_setting && subsecond % 2 && state->setting_state < alarm_setting_idx_pitch && !state->alarm_quick_ticks) {
         buf[_blink_idx[state->setting_state]] = buf[_blink_idx2[state->setting_state]] = ' ';
     }
     watch_display_string(buf, 0);
@@ -179,9 +180,9 @@ static void _alarm_indicate_beep(alarm_state_t *state) {
 
 static void _abort_quick_ticks(alarm_state_t *state) {
     // abort counting quick ticks
-    if (state->alarm_quick_ticks >= ALARM_QUICK_MIN_TICKS) state->alarm[state->alarm_idx].enabled = true;
-    if (state->alarm_quick_ticks >= 0) {
-        state->alarm_quick_ticks = -1;
+    if (state->alarm_quick_ticks) {
+        state->alarm[state->alarm_idx].enabled = true;
+        state->alarm_quick_ticks = false;
         movement_request_tick_frequency(4);
     }
 }
@@ -201,15 +202,12 @@ void alarm_face_setup(movement_settings_t *settings, uint8_t watch_face_index, v
             state->alarm[i].pitch = 1;
         }
         state->alarm_handled_minute = -1;
-        state->alarm_quick_ticks = -1;
     }
 }
 
 void alarm_face_activate(movement_settings_t *settings, void *context) {
     (void) settings;
     (void) context;
-    watch_display_string("  ", 8);
-    watch_clear_indicator(WATCH_INDICATOR_LAP); // may be unnecessary, but who knows
     watch_set_colon();
 }
 
@@ -219,7 +217,7 @@ void alarm_face_resign(movement_settings_t *settings, void *context) {
     _alarm_update_alarm_enabled(settings, state);
     watch_set_led_off();
     watch_store_backup_data(settings->reg, 0);
-    state->alarm_quick_ticks = -1;
+    state->alarm_quick_ticks = false;
     movement_request_tick_frequency(1);
 }
 
@@ -263,37 +261,17 @@ bool alarm_face_loop(movement_event_t event, movement_settings_t *settings, void
 
     switch (event.event_type) {
     case EVENT_TICK:
-        if (state->alarm_quick_ticks >= 0) {
-            // we are counting ticks for the alarm button
-            if (state->setting_state == alarm_setting_idx_hour || state->setting_state == alarm_setting_idx_minute) {
-                if (state->alarm_quick_ticks < INT8_MAX) state->alarm_quick_ticks++;
-                if (state->alarm_quick_ticks == ALARM_QUICK_MIN_TICKS) {
-                    // initiate fast counting
-                    movement_request_tick_frequency(8);
-                } else if (state->alarm_quick_ticks > ALARM_QUICK_MIN_TICKS) {
-                    // fast count hours or minutes
-                    if (state->setting_state == alarm_setting_idx_hour)
+        if (state->alarm_quick_ticks) {
+            // we are in fast cycling mode
+            if (state->setting_state == alarm_setting_idx_hour) {
                         state->alarm[state->alarm_idx].hour = (state->alarm[state->alarm_idx].hour + 1) % 24;
-                    else if (state->setting_state == alarm_setting_idx_minute)
+            } else if (state->setting_state == alarm_setting_idx_minute) {
                         state->alarm[state->alarm_idx].minute = (state->alarm[state->alarm_idx].minute + 1) % 60;
-                    _alarm_face_draw(settings, state, event.subsecond);
-                    break;
-                }
-            } else {
-                _abort_quick_ticks(state);
-            }
+            } else _abort_quick_ticks(state);
         } else if (!state->is_setting) break; // no need to do anything when we are not in settings mode and no quick ticks are running
-        // otherwise fall through
+        // fall through
     case EVENT_ACTIVATE:
         _alarm_face_draw(settings, state, event.subsecond);
-        break;
-    case EVENT_LIGHT_BUTTON_DOWN:
-        break;
-    case EVENT_ALARM_BUTTON_DOWN:
-        // check if we need to start counting ticks
-        if (state->is_setting && (state->setting_state == alarm_setting_idx_hour || state->setting_state == alarm_setting_idx_minute)) {
-            state->alarm_quick_ticks = 0;
-        }
         break;
     case EVENT_LIGHT_BUTTON_UP:
         if (!state->is_setting) {
@@ -372,14 +350,21 @@ bool alarm_face_loop(movement_event_t event, movement_settings_t *settings, void
                 break;
             case alarm_setting_idx_minute:
             case alarm_setting_idx_hour:
-                // hour or minute selection
-                _abort_quick_ticks(state);
+                // initiate fast cycling for hour or minute settings
+                movement_request_tick_frequency(8);
+                state->alarm_quick_ticks = true;
                 break;
             default:
                 break;
             }
         }
         _alarm_face_draw(settings, state, event.subsecond);
+        break;
+    case EVENT_ALARM_LONG_UP:
+        if (state->is_setting) {
+            if (state->setting_state == alarm_setting_idx_hour || state->setting_state == alarm_setting_idx_minute)
+                _abort_quick_ticks(state);
+        }
         break;
     case EVENT_BACKGROUND_TASK:
         // play alarm
@@ -405,7 +390,6 @@ bool alarm_face_loop(movement_event_t event, movement_settings_t *settings, void
     case EVENT_TIMEOUT:
         movement_move_to_face(0);
         break;
-    case EVENT_LOW_ENERGY_UPDATE:
     default:
       break;
     }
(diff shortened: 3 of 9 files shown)


movement/movement.h
```diff
@@ -61,7 +61,8 @@ typedef union {
         // altimeter to display feet or meters as easily as it tells a thermometer to display degrees in F or C.
         bool clock_mode_24h : 1;            // indicates whether clock should use 12 or 24 hour mode.
         bool use_imperial_units : 1;        // indicates whether to use metric units (the default) or imperial.
-        uint8_t reserved : 7;               // room for more preferences if needed.
+        bool alarm_enabled : 1;             // indicates wheter there is at least one alarm enabled.
```
wryun, line 64: Typo: wheter.

It would be good if multiple faces were able to interact on this since it's in settings (personally, I would like countdown to play an alarm as well, and it uses an indicator on its own face), but I can't think of a particularly nice way to do it in the current movement architecture (since who is responsible for turning the indicator off?). One simple change would be to use more bits here, and then you could maintain a counter (i.e. +1 for each alarm, -1 for each alarm turned off).

If we were refactoring more (probably not in this PR?), a different approach would be to maintain the alarm logic in movement.c (e.g. movement_schedule_alarm(time, rounds, alarm_note) or movement_schedule_background_task(time, alarm_indicator=true)), which would separate things a bit more cleanly. It wouldn't handle repeating alarms; you'd just do another movement_schedule_background_task when the previous one fired.

movement/watch_faces/complication/alarm_face.h
```diff
@@ -0,0 +1,79 @@
...
+} alarm_setting_t;
+
+typedef struct {
+    bool is_setting : 1;
+    uint8_t setting_state : 3;
+    uint8_t alarm_idx : 4;
+    uint8_t alarm_playing_idx : 4;
+    int8_t alarm_handled_minute;
+    alarm_setting_t alarm[ALARM_ALARMS];
+} alarm_state_t;
```
wryun, line 62: Thought (for another PR, probably): this is complex enough that it might be nice to persist to disk so it's not as easily lost.

movement/watch_faces/complication/alarm_face.c
```diff
@@ -0,0 +1,309 @@
...
+        - MF = the alarm fires Mondays to Fridays (yes, I am aware of the ambiguity of the abbreviation  ;-)
+        - WN = the alarm fires on weekends (Sa/Su)
+        - MO to SU = the alarm fires only on the given day of week
+    - You can fast jump (kinda) through hour or minute setting via long press of the alarm button.
+    - You can select the tone in which the alarm is played. (Three pitch levels available)
+    - You can select how many "beep rounds" are played for each alarm (1 to 9).
+    - The simple watch face indicates any alarm set by showing the bell indicator.
+*/
+
+static const char _dow_strings[ALARM_DAY_STATES + 1][3] ={"AL", "MO", "TU", "WE", "TH", "FR", "SA", "SO", "ED", "1t", "MF", "WN"};
```
wryun, line 55: Nit: Instinctively, I feel like not 'wasting' the NULL bytes here :) (i.e. [2] rather than [3])

movement/watch_faces/complication/alarm_face.c
```diff
@@ -0,0 +1,309 @@
...
+
+    Usage:
+    - In normal mode, the alarm button cycles through all 10 alarms. 
+    - Long pressing the alarm button in normal mode toggles the corresponding alarm on or off.
+    - Pressing the light button enters setting mode and cycles through the settings of each alarm.
+    - In setting mode an alarm slot is selected by pressing the alarm button when the slot number 
+      in the upper right corner is blinking.
+    - For each alarm slot, you can select the day. These are the day modes:
+        - ED = the alarm rings every day
+        - 1t = the alarm fires only one time and is erased afterwards
```
wryun, line 45: 👍

movement/watch_faces/complication/alarm_face.c
```diff
@@ -0,0 +1,309 @@
...
+    // don't show beep rounds in normal mode to avoid user confusion
+    if (!state->is_setting) 
+        buf[_blink_idx[5]] = buf[_blink_idx2[5]] = ' ';
+    // blink items if in settings mode
+    if (state->is_setting && subsecond % 2) {
+        buf[_blink_idx[state->setting_state]] = buf[_blink_idx2[state->setting_state]] = ' ';
+    }
+    watch_display_string(buf, 0);
+    
+    // draw pith level indicator
```
wryun, line 103: Typo: pitch

movement/watch_faces/complication/alarm_face.c
```diff
@@ -0,0 +1,309 @@
...
+        if (state->is_setting) {
+            _alarm_resume_setting(settings, state, event.subsecond);
+        }
+        break;
+    case EVENT_ALARM_BUTTON_UP:
+        if (!state->is_setting) {
+            // cycle through the alarms
+            state->alarm_idx = (state->alarm_idx + 1) % (ALARM_ALARMS);
+        } else {
+            // TODO: Handle the settings behaviour
```
wryun, line 232: Done? :)

wryun (commented):
Looks great! A few thoughts; haven't had a chance to try it out yet.

(and to be clear, I'm a random, not someone who has any control over the repo, so take any comments with a grain of salt :) )

movement/watch_faces/complication/alarm_face.h
```diff
@@ -0,0 +1,79 @@
...
+#ifndef ALARM_FACE_H_
+#define ALARM_FACE_H_
+
+#include "movement.h"
+
+/*
+A face for setting various alarms
+*/
+
+#define ALARM_ALARMS 10      // no of available alarm slots (be aware: only 4 bits reserved for this value in struct below)
```
wryun, line 36: I'd sort of like to have this user configurable, but can't think of a nice way to do it.

movement/movement.h
TheOnePerson: Thanks a lot for the detailed review!

> Typo: wheter

Will be fixed in next commit.

> ... One simple change would be to use more bits here, and then you could maintain a counter (i.e. +1 for each alarm, -1 for each alarm turned off).

This seems like a good idea at first glance, but I'm a bit conflicted about this. On the one hand, a counter like this would solve situations where you have more than one watch face wanting to use the bell indicator on the clock face. To implement this, you would probably need at least 4 bits for an alarm counter. It would be helpful (and to prevent overflows) to add a dedicated function in `movement.c` taking care of incrementing and decrementing the alarm counter, e.g. `movement_add_alarm(int8_t number_of_alarms)`. 
Also: If you would want to be able to not only _show_ if alarms are set on the clock face, but also _control_ if they fire (like on the original Casio firmware), you would still need the additional `alarm_enabled` flag.
This, on the other hand, feels like eating up a good chunk of reserved bits and adding complexity to movement.c for very specific use case functionality. Maybe @josecastillo has an opinion on this?

> If we were refactoring more (probably not in this PR?), a different approach would be to maintain the alarm logic in movement.c (e.g. movement_schedule_alarm(time, rounds, alarm_note) or movement_schedule_background_task(time, alarm_indicator=true)), which would separate things a bit more cleanly. It wouldn't handle repeating alarms; you'd just do another movement_schedule_background_task when the previous one fired.

If I am not mistaken the `movement_schedule_background_task(time, alarm_indicator=true))` approach would not really be suitable for alarms that can be scheduled for an arbitrary point in time. @joeycastillo wrote this on discord:

> If you implement the _wants_background_task function in your watch face, you'll get that call once a minute whether in sleep mode or wake mode. this is the preferred method for implementing alarms that fire on the minute mark, or scheduled tasks like a sensor reading or calculation. If your watch face schedules a background task with movement_schedule_background_task, the watch will remain in wake mode until the task fires, so that it can check every second for a match. this is the preferred method for timers or countdowns that may not fire exactly at the top of the minute.

movement/watch_faces/complication/alarm_face.h
TheOnePerson: Definitely a good proposal! But for now, I would have no idea on how to implement this. 😄 So yeah, rather an improvement for another PR.

movement/watch_faces/complication/alarm_face.c
TheOnePerson: Fix in next commit

movement/watch_faces/complication/alarm_face.c
TheOnePerson: Yes, done. ;-) Fix in next commit

movement/watch_faces/complication/alarm_face.h
TheOnePerson: Same here ;-)

movement/watch_faces/complication/alarm_face.c
TheOnePerson: You are right, this feels like a waste of 12 NULL bytes here. I just could not think of  a way of solving this without adding at least 12 additional bytes of code. (Please be kind here: it is literally decades ago since I have coded something in good old C - but it is a lot of fun, though! 😃)

TheOnePerson:
I converted this pr to draft now, because maybe there will be some refactoring regarding who the lcd is updated.

movement/watch_faces/complication/alarm_face.c
TheOnePerson: Fix is fixed in commit c65b1cc. I now use chars here instead of strings.

TheOnePerson:
The latest commit c65b1cc adds some more functionality:

- Number of alarm slots has been increased to 16.
- Alarms can not only have 1 to 9 regular alarm beeping rounds, but also be extra short (visualized by an 'o') resulting in two short beeps, and extra long, resulting in 20 beeping rounds (about 20 secs in total).

I would regard the alarm face feature-complete with this. Will try it out on my watch and see if everything goes smoothly.

movement/movement.h
wryun: > If I am not mistaken the movement_schedule_background_task(time, alarm_indicator=true)) approach would not really be suitable for alarms that can be scheduled for an arbitrary point in time.

Sorry, what I forgot to mention (because apparently it was all going on in my head!) is that it would be reasonably easy to change movement_schedule_background_task to check on the sleep mode loop as well, so that what you would have is 1 minute granularity rather than 1 second granularity once the watch has gone to sleep. TBH, I think the current scheduled task implementation is quite prone to footguns, because if ever you register a scheduled task and sleep before it executes, you can get it stuck in there forever (I think, @joeycastillo ?).

joeycastillo:
Thanks for opening this PR! I'm eager to review it once it's out of draft state, but this week I need to focus on kitting and shipping the next 280 boards. Once it's ready I will aim to get to it as soon as I can.

joeycastillo:
Let's hang tight on anything related to settings persistence. I want to slow walk this because if we get it wrong and folks end up with stale or corrupt configuration data on their watch, it will become very difficult to debug. Let's wait, think it through at a system-wide level and do it right.

TheOnePerson:
This sounds very reasonable, fair enough. 👍

wryun:
I've installed this to my watch now... but only with 4 alarms (16 just seemed a few too many for me...). I really like the interface - very slick! I'll see if it wakes me up tomorrow :)

wryun:
Alarm went off. I have it set currently to only go off on Thursdays, but because there's _an_ alarm active my indicator is on. I don't find this very useful - the next alarm is a long way away!

@TheOnePerson , what do you think about only showing the indicator if an alarm will happen in the next 24 hours? Or maybe 18 hours? This helps address the problem (for me) of 'have I set an alarm for tomorrow morning'.

TheOnePerson:
Hm - I'm not too convinced that indicating an alarm firing off within the next x hours is a no-brainer, @wryun. I personally like to see if there are _any_ alarms set or if the watch gives me peace, because no alarms are active. In a scenario where you have set multiple alarms for different days the indicator is not be very informative, admittedly, and it could be more helpful to see if there are alarms set for the next day. Will give this some more thoughts...

TheOnePerson:
There is another (small) thing nagging me: The simple watch face mixes up the indicators. On the original F91, the small bell icon represents the hourly chime and the bar-like symbol stands for an active alarm. However, the simple watch face uses the bigger bar-like symbol for the hourly chime. (And thus, I have implemented the bell icon as an indicator for active alarms.)
My guess would be that there is no specific reason for this particular use of symbols and I am thinking about switching this to be in line with the standard Casio usage.

wryun:
Yes, I'm not claiming my use is necessarily true for everyone; there are definitely reasons to prefer the existing implementation. I think it comes down to 'are you worried an alarm is _not_ set' (i.e. show indicator only for next morning's alarm) vs 'are you worried that an alarm _is_ set' (show indicator for any alarm).

If people are heavily using the specific day/weekday/weekend functions, I think they'll end up with the alarm indicator on all the time. i.e. it won't have much informative power for them. In my case, I know I always need to wake up early on Thursday, and might set once off alarms for 'special occasions'; what feels useful is to know that I've set my 'once off alarm'.

Re the symbol switch, I guess part of the idea is to use the bar alarm indicator for other none alerting things (e.g. it's used for temp/voltage reads).

TheOnePerson:
I just added the option to initiate a proper quick cycling through hour and minute setting by long pressing the alarm button.

TheOnePerson:
I guess the pr is ready for review now. There may be some discussions regarding the behavior or the alarm indicator on the simple watch face, and also the indicators (bell & signal icon) do not match the original Casio usage, but from a technical point of view, I would regard this alarm face ready for merge now.

joeycastillo:
Awesome! I will aim to do a proper review tomorrow, as I'm headed out of town today. I've been following this conversation and I have a few notes, none of which require immediate action: 

First: regarding the `alarm_enabled` flag and the idea of multiple watch faces interacting with it, I think it's reasonable to keep it a boolean, and to tell folks that they should generally only build Movement with one watch face for setting alarms. For example, the "Wake Face" is a watch face that sets one daily alarm, ala the original F-91W. It would be reasonable to say "Build your firmware with the wake face OR the advanced 16-alarm watch face, but not both.” This way any watch face that sets alarms can have full ownership of that flag. (I'm assuming in here that watch faces like the countdown timer wouldn't interact with the alarm indicator, only ones that schedule alarms). 

Second: with regard to the question of when the indicator should light — should it be on on Monday if the only alarm is set for Thursday — I think that if we say as above that your watch face has total control of the alarm flag, we should leave that up to your watch face. You could keep it on if any alarm is set, or use a background task to see if an alarm is set for the next 24 hours and only light the indicator in that case. Either way, knowing that your alarm face is the only one in control of that flag means you can set the flag however you see fit. 

Finally: regarding the mixup with the indicators, I'm down to swap them to match the original F-91W's usage, but that's something I can do in a separate PR. No idea how I got that so backwards for so long!

movement/watch_faces/complication/alarm_face.c
```diff
@@ -0,0 +1,414 @@
...
+        state->alarm_handled_minute = -1;
+        state->alarm_quick_ticks = -1;
+    }
+}
+
+void alarm_face_activate(movement_settings_t *settings, void *context) {
+    (void) settings;
+    (void) context;
+    watch_display_string("  ", 8);
+    watch_clear_indicator(WATCH_INDICATOR_LAP); // may be unnecessary, but who knows
```
joeycastillo, line 212: This should be unnecessary as the display is cleared before your watch face becomes active.

movement/watch_faces/complication/alarm_face.c
```diff
@@ -0,0 +1,414 @@
...
+                    else if (state->setting_state == alarm_setting_idx_minute)
+                        state->alarm[state->alarm_idx].minute = (state->alarm[state->alarm_idx].minute + 1) % 60;
+                    _alarm_face_draw(settings, state, event.subsecond);
+                    break;
+                }
+            } else {
+                _abort_quick_ticks(state);
+            }
+        } else if (!state->is_setting) break; // no need to do anything when we are not in settings mode and no quick ticks are running
+        // otherwise fall through
```
joeycastillo, line 286: Minor tweak: if you remove the "otherwise" (so the comment reads exactly `// fall through`), it will silence a gcc compiler warning that's currently showing up.

movement/movement_config.h
```diff
@@ -30,6 +30,7 @@
 const watch_face_t watch_faces[] = {
     simple_clock_face,
     world_clock_face,
+    alarm_face,
```
joeycastillo, line 33: Let's just make sure to remove this before merging so as not to change the default build :)

joeycastillo (commented):
This is looking good! I'm wearing the watch face now and will test some more in the coming days, but at first glance I just noticed a couple of issues which I've commented on inline.

movement/watch_faces/complication/alarm_face.c
```diff
@@ -0,0 +1,414 @@
...
+    }
+    //handle am/pm for hour display
+    uint8_t h = state->alarm[state->alarm_idx].hour;
+    if (!settings->bit.clock_mode_24h) {
+        if (h > 12) {
+            watch_set_indicator(WATCH_INDICATOR_PM);
+            h -= 12;
+        } else {
+            watch_clear_indicator(WATCH_INDICATOR_PM);
+        }
```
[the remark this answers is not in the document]
TheOnePerson: Ah okay, I will correct this in the next commit. (I'm from a "24h country", so I wasn't aware of the details about am/pm at noon/midnight.) ¯\_(ツ)_/¯

movement/watch_faces/complication/alarm_face.c
TheOnePerson: Good to know, will be eliminated.

movement/watch_faces/complication/alarm_face.c
TheOnePerson: My compiler does not throw a warning there, but thanks for the hint. Will correct this.

movement/movement_config.h
TheOnePerson: Sure

joeycastillo:
Thanks for making these changes! Testing on wrist has gone well, so I'm going to merge this in now so that I can look at the long press PR; since it included some alarm code, it seems merging this in will make it easier to review that PR.

PR #101, opened 2022-10-17
Adds a rate meter watchface

dps:
This pull request adds a rate meter watchface.

The rate meter shows the rate per minute at which the alarm button is being pressed. This is particularly useful in sports where cadence tracking is useful. For instance, rowing coaches often use a dedicated rate meter - clicking the rate button each time the crew puts their oars in the water to see the rate (strokes per minute) on the rate meter.%

### DIFF
--- a/movement/watch_faces/complication/ratemeter_face.c
+++ b/movement/watch_faces/complication/ratemeter_face.c
@@ -1,0 +1,97 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2022 Joey Castillo
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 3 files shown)


movement/watch_faces/complication/ratemeter_face.c
```diff
@@ -0,0 +1,97 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2022 Joey Castillo
```
joeycastillo, line 4: Could you change the name here to reflect your name? Just want to make sure that individuals' contributions are attributed properly (and that you're down with licensing yout watch face under MIT terms).

dps: updated - thanks! PTAL

PR #102, opened 2022-10-18
Authenticity joke

wryun:
Holding the alarm button down on the simple clock face displays CASIO, then animates to SENSOR, then animates to HACKED, then all normal display is flipped (read the digits upside down, but in original direction - i.e. RTL).

Can reverse flip by holding down again up until CASIO (but not all the way to HACKED).


wryun:
Sort of legit, but I didn't bother documenting my library function :)

watch-library/shared/watch/watch_private_display.c
```diff
@@ -87,10 +88,60 @@ void watch_display_character(uint8_t character, uint8_t position) {
...
+            continue;
+        }
+
+        for (int bit_pos = 0; bit_pos < 8; ++bit_pos) {
+            int bit = 1 << bit_pos;
+
+            if ((old_segdata[i] & bit) != (new_segdata[i] & bit)) {
+                old_segdata[i] ^= bit;
+
+                uint64_t segmap = Segment_Map[i] >> (8 * bit_pos);
```
wryun, line 131: Should probably pull this back to the start of the loop, as in the normal display. Originally, I was choosing which bits to flip on a different basis, but left->right actually looked the best (IMO).

watch-library/shared/watch/watch_private_display.c
```diff
@@ -87,10 +88,60 @@ void watch_display_character(uint8_t character, uint8_t position) {
...
+                uint8_t com = (segmap & 0xFF) >> 6;
+                uint8_t seg = segmap & 0x3F;
+                if (old_segdata[i] & bit) {
+                    watch_set_pixel(com, seg);
+                } else {
+                    watch_clear_pixel(com, seg);
+                }
+
+                // Delay after changing, so user can appreciate wonder.
+                delay_ms(100);
```
wryun, line 141: (and yes, this rudely locks things up, so maybe this is unmergeable... I feel like I've seen watch faces that aren't resilient to missing ticks)

watch-library/shared/watch/watch_private_display.c
```diff
@@ -64,14 +64,15 @@ void watch_display_character(uint8_t character, uint8_t position) {
     } else {
         if (character == 'R') character = 'r'; // R needs to be lowercase almost everywhere
     }
-    if (position == 0) {
-        watch_clear_pixel(0, 15); // clear funky ninth segment
-    } else {
+    if (position != 0) {
```
wryun, line 67: A bit confused here - why was this pixel cleared but not the other special pixels?

joeycastillo: There are 8 bits in a byte. This means we can map up to 8 segments per digit "automagically" using the [character set](https://github.com/joeycastillo/Sensor-Watch/blob/5f0ebd8c66b3512d67efc8174c5f65a86488a4ea/watch-library/shared/watch/watch_private_display.h#L30) and [segment map](https://github.com/joeycastillo/Sensor-Watch/blob/5f0ebd8c66b3512d67efc8174c5f65a86488a4ea/watch-library/shared/watch/watch_private_display.h#L129) data. Most digits have only seven segments, but two have more than seven: the weekday digits in position 0 and 1. This makes the eighth bit useful for controlling an extra segment, where available. In position 0, we use the eighth bit to control the center downstroke of the W or M. In position 2, we use that eighth bit to control the extra segment at the top left that creates the `T` in **ST**opwatch. 

Writing any character to these two positions automagically sets and clears the "extra" eighth segment, because they are controlled by that eighth bit. Alas, position 0 has a _ninth_ segment at the bottom left, and we can't automagically control it with the segment mapping code because there aren't nine bits in a byte. So we manually clean it up; it's the only segment in the text display that we can't control with the eight bits per character we have available.

wryun: Yup, I think I got this, I was just confused bywatch_set_pixel(0, 12) and watch_set_pixel(1, 12) below since they didn't have corresponding clears. But having now looked at https://joeycastillo.github.io/Sensor-Watch-Documentation/segmap.html, I can see that these are quite... strange segments :) Thanks!

watch-library/shared/watch/watch_private_display.c
```diff
@@ -87,10 +88,60 @@ void watch_display_character(uint8_t character, uint8_t position) {
...
+                    watch_set_pixel(com, seg);
+                } else {
+                    watch_clear_pixel(com, seg);
+                }
+
+                // Delay after changing, so user can appreciate wonder.
+                delay_ms(100);
+            }
+        }
+    }
```
wryun, line 144: Should add a watch_display_string at the end with the new data for the 'bonus' set_pixels.

watch-library/shared/watch/watch_private_display.c
wryun: Apparently this causes missed events as well (i.e. if you release the button while the animation is happening, you get nothing).

I guess this needs to be changed to have a segment data thing passed in some we can maintain a state, and then up the ticks so we can just keep calling morph. Maybe.

watch-library/shared/watch/watch_slcd.h
```diff
@@ -84,6 +84,10 @@ void watch_clear_display(void);
   */
 void watch_display_string(char *string, uint8_t position);
 
+bool watch_display_string_morph(char *old_string, char *new_string);
```
wryun, line 87: TODO: document (and below).

wryun:
My current plan for this is:

- do some more cleanups to the watch display stuff
- add a proper state machine
- either use a define to enable it, or simply have an entirely different face
- make the sequence of displayed words set via a const array (prob default to [SENSOR, HACKED])

joeycastillo:
I'm so sorry I've dawdled on reviewing this PR, but I have a few concerns. 

First I'm sad to say that I really don't think I can merge anything that puts the word Casio on the screen. I get that this is a joke, but Casio has lawyers and lawyers aren't paid for their sense of humor. This would be using a trademarked term in a way that authentic devices use to signal authenticity, and while I get that you're using it to joke about inauthenticity, I fear that the cease-and-desist letter would come to me. 😬

If we resolve that and decide on a different word (even just “HACKED”), I would definitely prefer to do this via a different watch face, just because I'm hesitant to bake an easter egg into the default watch face that ships with everyone's board. If someone wants to opt in to the easter egg that's great, but if we put it in by default, it's an additional support burden to either document the behavior so it's known, or explain the behavior to folks who encounter it unexpectedly. Especially if it causes the characters to become inverted. When I was testing that freaked me out, and I fear folks would see that as their watch being broken if they're not in on the joke.

Finally: I'm really hesitant about putting the `watch_display_string_morph` function in `watch_private_display.c`. It's just that that's the lowest level of the support libraries, below Movement itself, and I'd like the stuff that lives way down there to be as simple as possible. Having said that I do like your idea of splitting `watch_convert_char_to_segdata` out of `watch_display_character`; I would be open to making that a public API, so that you could move the `display_string_morph` function into your watch face's code, fetch the segment data from `watch_convert_char_to_segdata`, and perform the animation higher up in the stack.

Hope this helps and again I'm sorry for my trepidation; the morph effect looks hella cool, but the more core to the user and developer experience a thing is, the more I want to favor clarity and simplicity over the cool factor.

wryun:
:+1: all these are reasonable concerns. I'll look into cleaning this up in the way you've suggested.

wryun:
So, I've pushed it a bit more in the direction you suggested, but haven't yet moved it to a completely separate face, as I thought I'd run the '#ifdef' version past you. Please don't feel like you have to review this with urgency, or at all! I will not be offended :)

Issues:

- I didn't want to make a copy of simple_clock_face with the easter egg because I'd prefer not to have to keep it up to date with clock face changes. I've instead tried to make the '#ifdef' as clean a cut as I could.
- In some ways, this might be nice as an entirely separate face (easter_egg_face?) which could be entered from the simple_clock_face if it existed, but this would entail a bunch of unnecessary refactoring.
- I've moved morph out at your suggestion, but I'm not sure I like it. To avoid knowing too much about the details of segdata, I've made not only convert_segdata but also display_segment public. Regardless, this feels like very low level stuff. I could just make a copy of display_segment, but then I feel like I'm just demonstrating that we really shouldn't have access to segdata (at least keeping display_segment out I don't have to know the full details...).
- I've left display_invert in, because, well, there's no way to do that other than at a low level. The gist of your comment seemed to be that I should just remove the display inversion, but since you didn't explicitly say so ;)

movement/watch_faces/clock/simple_clock_face.c
```diff
@@ -27,6 +27,18 @@
 #include "watch.h"
 #include "watch_utility.h"
 
+// Defining the EASTER_EGG means that very long presses on 'ALARM' on this face
+// show these words on the screen, and once the end is reached the watch display
+// inverts. If EASTER_EGG is not defined, all of this code path is not included.
+// If there's more than one string in the easter egg array, after completion
+// the display will 'flip' all the characters. To undo this, start the easter
+// egg again (but don't let it complete).
+#define EASTER_EGG {"    SENSOR ", "    HACKED"}
```
wryun, line 36: [] disable before merge

watch-library/shared/watch/watch_slcd.h
```diff
@@ -84,6 +84,12 @@ void watch_clear_display(void);
   */
 void watch_display_string(char *string, uint8_t position);
 
+void watch_display_segment(uint8_t position, uint8_t bit_pos, bool on);
```
wryun, line 87: - [ ] document before merge

wryun:
I still run this on my watch, but every now and again it seems to hang with hacked on the screen and I need to press button to get out. Almost a feature, right? I think this is to do with some weird interaction between accidentally holding the button down (it's the one close to the wrist...) and sleep mode.

Also, the seconds no longer invert on the clock since I last updated from main. Some optimisation that I haven't actually looked into, but probably an easy fix.

joeycastillo:
Heyo! So I'm trying to tame the pull request situation and I'm afraid I'm going to have to close this one. Playing with the authenticity mark feels too spicy for my taste. Still it looks great, it's rad that you did it, and thanks for contributing! I'm afraid I just can't put it in main.

wryun:
Aw :( But it doesn't say Casio any more, just sensor!

Is it worth raising the refactor of the display code as a separate PR, as this will make it easier to carry my commit in my own repo?

WesleyAC:
@wryun one thought i had when i was working on #321 is that it should be possible to make a "proxy" face that calls out to another face except in a particular case — giving that a shot might make it easier to maintain, and would also mean people could use it on any face, not just the default clock face.

Still would be up to @joeycastillo, i can understand not wanting to merge something related to the authenticity function, but perhaps something more generic than that would be acceptable, and that approach could be useful for some other things as well, even if it doesn't get merged as something that looks like this (i'm imagining a face that lets people rebind buttons on any other face, for instance, or more complex navigation UIs built as a single face to make them easier to iterate on)

PR #103, opened 2022-10-18
Long press of mode on home screen goes to settings

wryun:
This also changes the default rotation so you don't see the settings screens.

It's an optional feature based on whether MOVEMENT_SECONDARY_FACE_INDEX is defined in movement_config.h. I could have written it to use ifdefs everywhere, but ifdefs are kinda ugly...

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -50,6 +50,11 @@
 #include "alt_fw/deep_space_now.h"
 #endif
 
+// Default to no secondary face behaviour.
+#ifndef MOVEMENT_SECONDARY_FACE_INDEX
+#define MOVEMENT_SECONDARY_FACE_INDEX MOVEMENT_NUM_FACES
+#endif
+
 #if __EMSCRIPTEN__
 #include <emscripten.h>
 #endif
@@ -203,7 +208,8 @@ void movement_move_to_face(uint8_t watch_face_index) {
 }
 
 void movement_move_to_next_face(void) {
-    movement_move_to_face((movement_state.current_watch_face + 1) % MOVEMENT_NUM_FACES);
+    uint16_t face_max = movement_state.current_watch_face < ((int16_t)MOVEMENT_SECONDARY_FACE_INDEX) ? MOVEMENT_SECONDARY_FACE_INDEX : MOVEMENT_NUM_FACES;
+    movement_move_to_face((movement_state.current_watch_face + 1) % face_max);
 }
 
 void movement_schedule_background_task(watch_date_time date_time) {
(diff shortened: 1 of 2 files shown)


movement/movement.c
```diff
@@ -409,14 +415,18 @@ bool app_loop(void) {
...
         // Note that it's the face's responsibility to provide some way to get to the next face, so if EVENT_MODE_BUTTON_* is
         // used for face functionality EVENT_MODE_LONG_PRESS should probably be handled and next_face() triggered in the face
         // (which would effectively disable the normal 'long press to face 0' behaviour).
         if (event.event_type == EVENT_MODE_LONG_PRESS 
-            && movement_state.current_watch_face > 0 
             && !movement_state.watch_face_changed) {
-            movement_move_to_face(0);
+            if (movement_state.current_watch_face != 0) {
+                movement_move_to_face(0);
+            } else if (MOVEMENT_SECONDARY_FACE_INDEX != MOVEMENT_NUM_FACES) {
```
joeycastillo, line 427: Question: would it make more sense to `#define MOVEMENT_SECONDARY_FACE_INDEX 0` if not defined? This way the conditional could just be `else if (MOVEMENT_SECONDARY_FACE_INDEX)`. There's no scenario where the secondary face will be the first face, and it feels cleaner.

movement/movement_config.h
```diff
@@ -39,4 +39,11 @@ const watch_face_t watch_faces[] = {
 
 #define MOVEMENT_NUM_FACES (sizeof(watch_faces) / sizeof(watch_face_t))
 
+/* Determines what face to go to from the first face if you've already set 
+ * a mode long press to go to the first face in preferences, and
+ * excludes these faces from the normal rotation.
+ * Usually it makes sense to set this to the preferences face.
+ */
+#define MOVEMENT_SECONDARY_FACE_INDEX (MOVEMENT_NUM_FACES - 2)
```
joeycastillo, line 47: If we merge this, I would like to have this #defined to 0 by default. The standard F-91W has its time settings screen in rotation, and all the documentation describes settings and time set as always being available as the last two watch faces. Changing the default behavior to hide the settings screens represents a huge change to the UX.

joeycastillo (commented):
I'm hesitant about this change. For one thing, I think the Mode long press on the first watch face has a lot of potential, and I'm not certain that moving to the settings screen is the best use for it. I would hate to make this behavior standard, and then later realize there's a better use for this one remaining gesture. I also have concerns about skipping the settings screens being less discoverable and conflicting with documentation folks have already read and internalized.

Having said that, if this behavior is added but not set as the default, I'm open to merging.

joeycastillo:
To be clear, I like this a lot! It is cleaner to remove those faces from the carousel, the gesture does make sense, and I think your implementation is sound. I'm just conservative about changing the default behavior right now :)

wryun:
Thanks for making the change and merging.

PR #104, opened 2022-10-19
Make thermistor logging face show current temp.

wryun:
Use position 0 in the logging face to show the current temp (and echo the 'signal' behaviour from the readout face).

This means you don't need to 'waste' a face if you want logging and readout.

### DIFF
--- a/movement/watch_faces/sensor/thermistor_logging_face.c
+++ b/movement/watch_faces/sensor/thermistor_logging_face.c
@@ -28,20 +28,19 @@
 #include "thermistor_driver.h"
 #include "watch.h"
 
-static void _thermistor_logging_face_log_data(thermistor_logger_state_t *logger_state) {
+static void _thermistor_logging_face_update_data(thermistor_logger_state_t *logger_state) {
     thermistor_driver_enable();
     watch_date_time date_time = watch_rtc_get_date_time();
     size_t pos = logger_state->data_points % THERMISTOR_LOGGING_NUM_DATA_POINTS;
 
     logger_state->data[pos].timestamp.reg = date_time.reg;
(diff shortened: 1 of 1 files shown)


joeycastillo:
I'm hesitant about this change. First, omitting the temperature display face means there's no way to toggle between Celsius and Fahrenheit. And while you might be able to work a gesture into this face that does that, it still feels like we're trying to add a lot of complexity to what is already perhaps one of the most complex watch faces. 

More broadly, I like watch faces that do one thing and do it with clarity. Far from "wasting" a face, having one face that only does temperature display clears a lot of cognitive overhead for the wearer. They know that when the line at the top reads "TE", they are viewing the current temperature, and when it reads "TL" they are viewing the temperature log. There's no confusion about which gesture switches between several modes within one watch face (current temp, logged temp, timestamp of logged temp); the Mode button simply switches the watch face from "temperature" to "temperature log".

The [Watch Interface Guidelines](https://www.sensorwatch.net/docs/wig/) document (admittedly a work in progress) has some guidance on this, but the gist is that you can imagine a watch face being designed to answer one simple prompt quickly and clearly. To that end, I feel that “How cold is it right now” and “How cold was it last night” are two distinct prompts deserving of two separate watch faces.

wryun:
I think my preference would be for the 'use_imperial' setting to go in the preferences face. I guess the problem I'm trying to address is that I'm rarely interested in the temperature, but would like it there just in case without clogging things up. For me it's also clear in the sense that I think of '0' as the current state, and '1' as the previous hour, etc. When I initially used the TL face, I was surprised I couldn't access the current temperature; the way I would resolve the TL/TE oddness would be to remove the existing TE face and rename this one TE.

However, I'm totally okay with this being rejected as against the interface guidelines, so I'll close it off now :)

PR #105, opened 2022-10-20
Fix scheduled tasks scheduling another task

niclashoyer:
I stumbled on this while creating a watch face that needs a (kinda) periodic countdown.

The watch face handles the background task event and immediately schedules a new background task. Without these changes the next background task is not run if it was the only watch face with a background task, because `num_active_tasks` is zero.

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -159,6 +159,10 @@ static void _movement_handle_scheduled_tasks(void) {
                 scheduled_tasks[i].reg = 0;
                 movement_event_t background_event = { EVENT_BACKGROUND_TASK, 0 };
                 watch_faces[i].loop(background_event, &movement_state.settings, watch_face_contexts[i]);
+                // check if loop scheduled a new task
+                if (scheduled_tasks[i].reg) {
+                    num_active_tasks++;
+                }
             } else {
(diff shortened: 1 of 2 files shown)


joeycastillo:
This makes sense to me! I may need to fix one thing in the scheduled tasks API for this to work well though: currently watch faces can only set a scheduled task when they are active, because Movement sets up the background task at the index of the active watch face. So if a watch face tries to schedule a new task while it is off screen, this check won't succeed (because it'll be at the wrong index). 

I need to make `movement_schedule_background_task` and `movement_cancel_background_task` take the watch face index as a parameter. It was lazy API design on my part, and fixing it is something I'd like to do in this PR before merging.

joeycastillo:
I'll make these changes tomorrow and push to this branch.

joeycastillo:
OK, this should be good to go! In your sailing watch face, save (in your watch face context) the watch face index that you get in `sailing_face_setup`, and then use the new `movement_schedule_background_task_for_face` and `movement_cancel_background_task_for_face` functions, passing in your watch face index. This should (_should!_) allow your watch face to set new background tasks within a background task handler.

PR #106, opened 2022-10-21
Add sailing watch face

niclashoyer:
This is my first take at a "sailing" watch face.

Sailing races are usually started according to rule World Sailing [rule 26](https://www.racingrulesofsailing.org/rules?part_id=111). On the water visual and acoustic signals follow a special countdown:

- 5 minutes to start
- 4 minutes to start
- 1 minute to start
- start

This watch face is a modified countdown watch face that has 6 slots for countdowns with the following features:

 - rule 26 is the default, which is `5 4 1 0 0 0`. This is configurable in the settings. `0` means "start" and resets the countdown.
 - beep on each signal
 - pressing the alarm button "syncs" the countdown (useful if e.g. the 5 minute signal was missed to sync to the 4 minute signal)
 - long pressing the light button resets the countdown
 - last minute is displayed in large letters

However there are still some bugs:

 - [x] periodic background tasks (a task scheduling another task) needs #105 **RESOLVED**
 - [x] the watch face does not resign, but is unable to schedule a task in low power mode, which causes the countdown to be missed (e.g. while counting from 4 to 1 minute)
   is there any way around this? Maybe preventing low power mode? Resigning is not useful anyway, as having a visual countdown to start is crucial. I'm not really sure this is caused by the low power mode, but this only happens if the countdown is running for some minutes.
   **EDIT**: after some more testing, this seems to be an issue with the simulator. After investigation it seems that the simulator does not run properly in the background (e.g. if another tab in the browser is visible). See #107 .

Possible improvements:

 - documentation
 - add beeps for the last seconds before start
 - start a "stopwatch" like timer after start

### DIFF
--- a/movement/make/Makefile
+++ b/movement/make/Makefile
@@ -57,6 +57,7 @@ SRCS += \
   ../watch_faces/complication/totp_face.c \
   ../watch_faces/complication/sunrise_sunset_face.c \
   ../watch_faces/complication/countdown_face.c \
+  ../watch_faces/complication/sailing_face.c \
   ../watch_faces/complication/counter_face.c \
   ../watch_faces/complication/blinky_face.c \
   ../watch_faces/complication/moon_phase_face.c \
--- a/movement/watch_faces/complication/sailing_face.c
+++ b/movement/watch_faces/complication/sailing_face.c
@@ -1,0 +1,268 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2022 Wesley Ellis
+ * Copyright (c) 2022 Niclas Hoyer
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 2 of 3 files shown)


dutchmartin:
Just saw this as a fellow sailor and programmer and I just wanted to say awesome work! If i can help you out with something, just let me know.

niclashoyer:
@dutchmartin thanks for the heads up! I think this is pretty complete and working as it is. There are possible improvements (as mentioned). I think we just need a bit documentation and then this could go to the main branch 🚀

joeycastillo:
Thanks for this PR! I'm down to merge this now, and aim to documentation at a later date. 

Thanks for making this!

Hein-NonesensE:
Hi there, thanks for the face, nice work! I'm quite happy with it so far, some great ideas!

As both a sailor and a race official, I'd love the suggested beeps for the last 10 seconds (I would add another single beep at 30s and at every minute that is not set to double-beep) and stopwatch-like timer after start. Maybe even with a lap counter? Might be too confusing though. Stop counting at 24h would probably be sufficient, switching to days/hours/minutes seems overkill.

Other Ideas I had:

- Optional beep settings: All beeps, no beeps, beep only at start (to also serve as an ordinary countdown face).
- Face name: As the day digits are not in use, I'd suggest SA on the weekday and IL on the day digits

What do you think?

PR #109, opened 2022-10-23
Make movement auto fire long press event

TheOnePerson:
As discussed on Discord this PR changes the behavior of movement firmware: the *_LONG_PRESS events are fired automatically when a button is pressed for longer than 64 ticks (half a second). This makes the long press experience a lot snappier than the current implementation.

To avoid disturbing existing watch faces, a new event type is introduced to signal a button release after a long press: the EVENT_(*)_LONG_UP event. Watch faces can react to this, if they want.

Additionally, in the alarm watch face I had to adjust the quick cycling logic (for hours/minutes) to make use of the new behavior of the movement firmware. (The watch face logic is much simpler than it was before!)

Notice: In the movement files, there is also (again) some logic for customizable alarms. This is thanks to my lack of git wizardry... ;-/

Edit: Added some words about why this is useful.

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -22,6 +22,8 @@
  * SOFTWARE.
  */
 
+#define MOVEMENT_LONG_PRESS_TICKS 64
+
 #include <stdio.h>
 #include <string.h>
 #include <limits.h>
@@ -250,11 +252,16 @@ void movement_play_signal(void) {
 }
 
 void movement_play_alarm(void) {
+    movement_play_alarm_beeps(5, BUZZER_NOTE_C8);
+}
+
+void movement_play_alarm_beeps(uint8_t rounds, BuzzerNote alarm_note) {
+    if (rounds == 0) rounds = 1;
+    if (rounds > 20) rounds = 20;
     movement_request_wake();
-    // alarm length: 75 ticks short of 5 seconds, or 4.414 seconds:
-    // our tone is 0.375 seconds of beep and 0.625 of silence, repeated five times.
-    // so 4.375 + a few ticks to wake up from sleep mode.
-    movement_state.alarm_ticks = 128 * 5 - 75;
+    movement_state.alarm_note = alarm_note;
+    // our tone is 0.375 seconds of beep and 0.625 of silence, repeated as given.
+    movement_state.alarm_ticks = 128 * rounds - 75;
     _movement_enable_fast_tick_if_needed();
 }
 
@@ -453,10 +460,13 @@ bool app_loop(void) {
     if (movement_state.alarm_ticks >= 0) {
         uint8_t buzzer_phase = (movement_state.alarm_ticks + 80) % 128;
         if(buzzer_phase == 127) {
+            // failsafe: buzzer could have been disabled in the meantime
+            if (!watch_is_buzzer_or_led_enabled()) watch_enable_buzzer();
+            // play 4 beeps plus pause
             for(uint8_t i = 0; i < 4; i++) {
                 // TODO: This method of playing the buzzer blocks the UI while it's beeping.
                 // It might be better to time it with the fast tick.
-                watch_buzzer_play_note(BUZZER_NOTE_C8, (i != 3) ? 50 : 75);
+                watch_buzzer_play_note(movement_state.alarm_note, (i != 3) ? 50 : 75);
                 if (i != 3) watch_buzzer_play_note(BUZZER_NOTE_REST, 50);
             }
         }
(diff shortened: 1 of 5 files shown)


movement/movement.h
```diff
@@ -109,13 +110,16 @@ typedef enum {
     EVENT_TIMEOUT,              // Your watch face has been inactive for a while. You may want to resign, depending on your watch face's intended use case.
     EVENT_LIGHT_BUTTON_DOWN,    // The light button has been pressed, but not yet released.
     EVENT_LIGHT_BUTTON_UP,      // The light button was pressed and released.
-    EVENT_LIGHT_LONG_PRESS,     // The light button was held for >2 seconds, and released.
+    EVENT_LIGHT_LONG_PRESS,     // The light button was held for >2 seconds, but not yet released.
```
wryun, line 113: Comment should say 1/2 second, perhaps? (I know, this issue already existed)

movement/movement.h
```diff
@@ -109,13 +110,16 @@ typedef enum {
     EVENT_TIMEOUT,              // Your watch face has been inactive for a while. You may want to resign, depending on your watch face's intended use case.
     EVENT_LIGHT_BUTTON_DOWN,    // The light button has been pressed, but not yet released.
     EVENT_LIGHT_BUTTON_UP,      // The light button was pressed and released.
```
wryun, line 112: Perhaps comments on this should change, as AFAIK this event doesn't occur after a long press.

Ideally, it would be renamed EVENT_LIGHT_BUTTON_SHORT_UP I think (i.e. globally replace this in the codebase). And similarly for the others.

movement/movement.c
```diff
@@ -557,9 +567,21 @@ void cb_fast_tick(void) {
     movement_state.fast_ticks++;
     if (movement_state.light_ticks > 0) movement_state.light_ticks--;
     if (movement_state.alarm_ticks > 0) movement_state.alarm_ticks--;
+    // check timestamps and auto-fire the long-press events
+    // Notice: is it possible that two or more buttons have an identical timestamp? In this case
+    // only one of these buttons would receive the long press event. Don't bother for now...
+    if (movement_state.light_down_timestamp > 0)
+        if (movement_state.fast_ticks - movement_state.light_down_timestamp == MOVEMENT_LONG_PRESS_TICKS + 1)
```
wryun, line 574: I've a bit confused here. Where is light_down_timestamp being set/reset?

wryun (commented):
Just from reading; haven't tried yet.

movement/movement.h
TheOnePerson: As for the EVENT_LIGHT_BUTTON_SHORT_UP idea: I tried to avoid renaming existing events to assure compatibility with existing code. Maybe @joeycastillo some time decides to take this road (of using different names) to clear things up.

movement/movement.c
TheOnePerson: The function `cb_light_btn_interrupt(void)` initiates that by calling `_figure_out_button_event(pin_level, EVENT_LIGHT_BUTTON_DOWN, &movement_state.light_down_timestamp)`.

movement/movement.h
TheOnePerson: You are right, the >2 seconds comment seems to be outdated. Will be corrected in next commit.

movement/movement.h
wryun: I agree it would cause some friction, but because in theory the faces are ending up in the same codebase and it wouldn't require significant changes to any people's existing work I think it's ok. But as you say, definitely @joeycastillo 's call :)

joeycastillo:
It looks like my merge of the alarm face created some conflicts, I think it's mostly just choosing the changes from `main` (except for the `EVENT_ALARM_LONG_UP` case?) but @ThatOnePerson I might ask you to resolve these just or offer guidance, just to make sure I don't make it worse :)

TheOnePerson:
I have just resolved the conflicts. We should be good to go now.

joeycastillo:
Tested this for a bit and it's looking good! Tested with several watch faces that make use of the long press and all of them seem much improved by the new behavior. Only one required a change, the pulsometer, but switching to the new `EVENT_ALARM_LONG_UP` makes it work exactly as expected. 

Regarding the naming: heard. There is some inconsistency and I think renaming could make it clearer, but for now I prefer to keep the event names the same. A big change to nomenclature, while unlikely to break anything, is something I'd prefer to do in a more major release with better messaging around the change.

Merging this one, thanks y'all!

PR #110, opened 2022-10-24
Fix Linux Makefile parallelism

wryun:
Makefile != bourne shell, I think.

At least on my systems, this was causing a raw -j, which meant that make had no restriction on its parallelism
(i.e. tried to build everything at once).

Couldn't see how to trivially show exactly what was happening (i.e. why ``nroc`` was evaluating to the empty string (?) ), but at least this changed fixed the issue and seems harmless.

### DIFF
--- a/make.mk
+++ b/make.mk
@@ -31,7 +31,7 @@ else
 endif
 
 ifeq ($(DETECTED_OS), LINUX)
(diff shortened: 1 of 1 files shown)


joeycastillo:
Looks good to me! Passes tests and I also tested on MacOS. 

Thanks!

PR #111, opened 2022-10-27
Add calculator watch face

niclashoyer:
Well, I'm curious how far we can go with watch faces 🚀 

This adds a full blown calculator to movement. There are some hacks and shortcuts, but it basically works as it should.

## Features

 - Uses [RPN](https://en.wikipedia.org/wiki/Reverse_Polish_notation) to minimize needed user input and ease of implementation
 - 32 bit float arithmetic
 - variable stack size (currently 4, can be changed at compile time)
 - currently displays up to 4 whole number digits and two decimal digits

## Documentation

The calculator uses [RPN](https://en.wikipedia.org/wiki/Reverse_Polish_notation) notation.  So instead of using infix operators, the operators always follow the parameters.

Example: To calculate `4 - 3` one would enter `4 3 -`

Using this notation also does not require any braces, so instead of `2 * (3 - 2)` one would enter `2 3 2 - *`.

The parameters are put on to a stack, which currently has a size of 4. So if one enters `2 3 2 - *` the stack would contain `2 3 2 0`. The subtraction then uses the first two values from the stack and pushes the result back on the stack, which leaves `1 2 0 0`. The multiplication then uses the next two values from the stack and pushes the end result: `2 0 0 0`.

### Normal mode

In normal mode the top of the stack is displayed (initially zero), which usually represents the result of the calculation.

 - `ALARM` enters number mode
 - `LIGHT` enters operator mode
 - `MODE` switches to the next watch face

### Number mode

Number mode pushes a new parameter on the stack. The first four digits are whole numbers, the last two (smaller) digits are decimals.

- `LIGHT` cycles through digits
- `ALARM` increases the selected digit
-  `MODE` pushes the parameter on the stack (and goes back to normal mode)

### Operation mode

Operation mode executes operations on the stack. Parameters are taken from the stack and the result is pushed back. If there aren't enough parameters on the stack, the calculator will go into error mode.

The display shows the current selected operation.

 - `LIGHT` cycles through the available operations
 - `ALARM` executes the selected operation (and goes back to normal mode, which will display the result)

Currently implemented operations:

 - add (2 params)
 - sub (2 params)
 - mul (2 params)
 - div (2 params)
 - pow (2 params)
 - sqrt (1 param)
 - pi (0 params, will just push pi onto the stack)

### Error mode

An error has happened (currently only caused by too few parameters on the stack).

 - `ALARM` goes back to normal mode
 - `MODE` switches to the next watch face

## Example

To calculate the following equation:

```2 * (3 - 2)```

which would be the following in RPN notation:

```2 3 2 - *```

the following button presses are needed:

 - `ALARM` enter number mode
 - `ALARM` 1
 - `ALARM` 2
 - `MODE` push number, back to normal mode
 - `ALARM` enter number mode
 - `ALARM` 1
 - `ALARM` 2
 - `ALARM` 3
 - `MODE` push number, back to normal mode
 - `ALARM` enter number mode
 - `ALARM` 1
 - `ALARM` 2
 - `MODE` push number, back to normal mode 
 - `LIGHT` enter operation mode
 - `LIGHT` change operation to `sub`
 - `ALARM` execute `sub`, back to normal mode
 - `LIGHT` enter operation mode
 - `LIGHT` change operation to `mul`
 - `ALARM` execute `mul`, back to normal mode

## Todo

 - [ ] more operations (e.g. trigonometric functions), maybe add banks of operations that can be switched with `MODE` for easier access (e.g. basic, trigonometric, constants)
 - [ ] add indicator if the number is greater than 9999 and does not fit the display
 - [ ] option to display/enter six whole digits and six decimals (as an idea: long press alarm and show whole/decimal using AM/PM)
 - [x] proper handling of infinity and NaNs
 - [ ] remove debug output and FIXMEs


wryun:
@niclashoyer you made me make a PR - https://github.com/joeycastillo/Sensor-Watch/pull/112 - somewhat different approach. Let me know your thoughts! I've obviously gone a bit overboard with how weird it is to interact with, but I think it's faster to enter numbers. It also does some weird stuff to display floating point numbers that you might be interested in.

niclashoyer:
@wryun I'll check that out! I think some kind of weirdness will always be there, as the input and output capabilities are so limited. I think that may be personal preference in the end. I would be very interested in a `watch_display_float` function in movement. This could be handy for a lot of other stuff.

I hope we'll find a way to do some kind of linking/compiling in wasm, so that it may be possible to build a configurator that enables a custom build just from the browser. That way we don't have to build alternative fw's for everybody 😄

wryun:
@niclashoyer https://sensor-watch-builder.fly.dev/ (ctrl to select things, but uses a server and not the browser; the libs are too enormous IMO to ever get emscripten working in the browser in a sane way).

I've also cleaned up my PR so it's a little more usable, for me at least :)

niclashoyer:
@wryun wow this is so great! Yes! Do you maybe want to opensource this?

wryun:
It's open source, but it's... I'm going to call it 'experimental' at the moment: https://github.com/wryun/sensor-watch-builder/

https://github.com/wryun/sensor-watch-builder/blob/main/TODO

(I was also using it as an opportunity to learn about openresty/lua, so it probably has some odd code)

joeycastillo:
I'm reviewing this now and it works pretty well! I am seeing some weirdness with number input — specifically when less significant digits are set, changing the more significant digits seems to mess up the less significant digits? For example:

* Enter the number 123 in the last three digits
* Advance to the thousands place and attempt to make the number 5,123
* The least significant digit seems to go down as the most significant digit goes up?

There are a couple of other quirks that I'm not sure are bugs or the way the UI is intended to work, but overall it works well so far! I made one change, to the display of the multiply operator to make it clearer; other than that, would you be willing to rename the watch face `rpn_calculator_face`? Just to signal to folks that this calculator may not work the same way as a "traditional" calculator they might expect, and in case we implement a calculator with infix notation in the future.

niclashoyer:
Thanks for the review! I think the number entry has a bug, I'll need to check that again. I'm happy with renaming. I think I can get this ready before Christmas 🎁

niclashoyer:
@joeycastillo I fixed the issues. Can you please add your changes for `mul` again? I accidentally force pushed and now they are gone 🙈 

> There are a couple of other quirks that I'm not sure are bugs or the way the UI is intended to work

Can you explain this further? I might be adding more functions (e.g. `sin`, `cos`, `tan`) and maybe paging for operators, to group them in basic, trigonometric and constants. But I don't want risk to add more quirky behavior 😄

joeycastillo:
I think I get it now! I still have some issues with entering numbers in the decimal mode; for example:

* `ALARM` enter Number mode
* `ALARM` set value to 1
* `LIGHT` ×5 to move to the tenths place
* `ALARM` set value to 1.1
* `ALARM` set value to 1.2

I think a third press of `ALARM` should set the value to 1.3, but instead it sets it to 1.29. And now, if I press `LIGHT` repeatedly to get back to the hundredths place, repeated presses of `ALARM` can't get me past the value 1.26. I think this is a glitch in number entry? But I'm not sure. 

Having said that, I'm down to merge this in now if you want so we can have it in main. That will make it easier for folks to work on it, and we can address any issues or TODO items in a future PR. 

Let me know what you think, and thanks!

niclashoyer:
@joeycastillo 

yes it seems that this is a bug. `ALARM` should just increase the digit and wrap back to 0. But I did not came up with a really good solution to do this. [The current function](https://github.com/joeycastillo/Sensor-Watch/pull/111/files#diff-f22121154dc86840e705e1b184b044f37e79a0d5e344c584246b945d424d9a32R78-R95) is very hacky and uses string conversion. This might be a problem due to float aliasing.

But I think we can merge this and improve this together.

PR #112, opened 2022-10-28
RPN Calculator face

wryun:
Differences from https://github.com/joeycastillo/Sensor-Watch/pull/111:

- uses the top two chars for the operation (so you can see the operation and the top of the stack at the same time). ALARM changes the operation you want, LIGHT executes it.
- when you've chosen the 'No' (number) operation, you get to enter a number. This works in 'guess a number'/'binary search' fashion, where ALARM says 'I want a bigger number' and MODE says 'I want a smaller number'.
- if displaying numbers that require a decimal point, uses a colon for the decimal point and the 'day' numbers for the exponent bit


movement/watch_faces/complication/calculator_face.c
```diff
@@ -0,0 +1,263 @@
...
+    (void) settings;
+    if (*context_ptr == NULL) {
+        *context_ptr = malloc(sizeof(calculator_state_t));
+        memset(*context_ptr, 0, sizeof(calculator_state_t));
+        // Do any one-time tasks in here; the inside of this conditional happens only at boot.
+    }
+    // Do any pin or peripheral setup here; this will be called whenever the watch wakes from deep sleep.
+}
+
+static void calculator_face_display_number(double num) {
```
wryun, line 40: @joeycastillo how would you feel about a separate PR which put this function in the library? (i.e. watch_display_float)

niclashoyer:
@wryun I'll be honest here. I don't think the guess number like input is very comfortable 😅 But as I said in #111 this might all come down to personal preference, I guess.

I do like the fact that it always displays the top of the stack.

How would go about adding other operations like sqrt or sin/tan?

movement/watch_faces/complication/calculator_face.c
joeycastillo: Not opposed to it, tho I will say that I think watch faces may want to have different strategies for displaying numbers; I'd like to think through it before offering a library function for it. For example, the location setting screen displays 40.73 degrees west as `- 40₇₃`, with the smaller numbers in the seconds place representing the decimal portion, whereas the temperature display shows 72.3°F as `72-3°F` (since there's no decimal segment, and the underscore doesn't work in all positions). 

Open to proposals on how to generalize this, but my instinct has been to have the library focus on displaying strings, and leave it up to individual watch faces how they want to format numbers into those strings.

niclashoyer: After playing around with the various methods this function uses to display numbers I think @joeycastillo might be right. The display is imho too constrained to really generalize this. Every watch face should proper document how they display their numbers and as long as it makes sense to the user, I think this is okay 👍🏽

wryun: I totally agree it's not fully generalisable, but it still seemed like it might be a useful utility function that can handle displaying an arbitrary float. If you know more about the number (as in lat/long, temp) there's no reason to use it. Happy to keep it here, at any rate.

wryun:
@niclashoyer thanks for trying my 'bigger/smaller' number thing :) It works in my head, but maybe that's more a reflection of me than the technique. It is at least _faster_ I believe... well, once you know what you're doing.

I did ponder adding a morse code or a polybius square mode - or perhaps binary input? - but that seemed a little more opaque :laughing:

wryun:
In terms of other operations, there's already a function mapping table (i.e. it's a couple of lines to add each function).

Possibly, as you mentioned in your PR, you'd want a way to move between different groups though.

movement/watch_faces/complication/calculator_face.c
```diff
@@ -0,0 +1,345 @@
...
+        watch_display_string(buf, 2);
+        return;
+    }
+
+    sprintf(buf, "%2d%c%05d", exponent, negative ? '-' : ' ', (int)round(num * 10000));
+    watch_set_colon();
+    watch_display_string(buf, 2);
+}
+
+#define C (s->stack[s->stack_index])
```
wryun, line 115: This was a bit of an experiment at writing code the way I usually wouldn't :)

(i.e. using macros and making everything deal with the global state instead of having clean functions)

Because the scope is so small, I _think_ it reads better.

wryun:
I've cleaned it up and added a bunch more functions. I honestly find this surprisingly usable in the simulator, though I haven't tried it on a watch yet. Main open question for me is should I add a way to 'browse' the stack as an operation... probably too confusing.

wryun:
Minor functions to add to make stack manipulation easier:

- [ ] swap
- [ ] drop/pop
- [ ] clear

watch-library/shared/watch/watch_private_display.c
```diff
@@ -91,6 +91,27 @@ void watch_display_character(uint8_t character, uint8_t position) {
     if (character == 'T' && position == 1) watch_set_pixel(1, 12); // add descender
     else if (position == 0 && (character == 'B' || character == 'D')) watch_set_pixel(0, 15); // add funky ninth segment
     else if (position == 1 && (character == 'B' || character == 'D' || character == '@')) watch_set_pixel(0, 12); // add funky ninth segment
+    else if (position == 0) {
+        // Clean-up for fancy symbols in the first location.
```
wryun, line 95: This is definitely the wrong thing to do, I just wasn't sure what the right thing to do was.

In particular, I was concerned by this comment:

```
0b11000000, // * (The + sign for use in position 0) 
```

which suggests that some other face is relying on * looking like + in position 0.

I think the best solution is to have an entirely separate Character_Set map for position 0. Thoughts welcome.

(Well, the _real_ best solution would be to calculate appropriate ISDATA things for characters in every position and add some combining logic, I think, as this would make things super fast/easy)

joeycastillo: My thinking was that you would just emit whatever character lights up the correct segments; unless you need a character that doesn't exist in the character map, there's no need to go this deep into the watch library. 

It looks like you'd use `*` for addition here, and `H` for multiplication. I get where it's confusing, but it solves the 99% case of being able to display something resembling a + at every location, and having a special case for position 0. 

I get that it's a bit confusing for this case in particular, but I'd rather clear that up with a comment in the watch face — Pulsometer is the only one that currently uses a + in this position — than spend additional cycles on a special character set for this one position.

wryun: Sure, will do. Thanks! It doesn't fit too badly in the code.

With the division one, I prefer using the middle segment (i.e. skewed Z shape, rather than separated bars) - thoughts? Or does it cause other issues?

joeycastillo: Sorry for the delay in reviewing this. It's not an issue that your watch face wants to display it this way, but I'd rather not code the special cases into `watch_private_display`, which is a lot deeper than it needs to be. It's much better, if your watch face needs to display a particular symbol, to put these pixel writing bits into your watch face. 

Putting them here means every watch face needs to go through this switch statement every time they write a character to position 0, even if the watch face that needs them isn't installed.

joeycastillo:
I tried it out for a bit, and I like it! I especially love the use of the signal beep for an error condition; it just feels right. I must confess I also find it challenging to use the guessy / binary search style of number entry, but I agree that it's likely a matter of taste; no reason not to have two options!

Similar to my feedback in the other PR, I would like to rename this face `rpn_calculator_alternate`, just to signal to folks that it's not an infix-notation calculator, and that it uses an alternate input method than they may expect from other watch faces.

If we can resolve that naming issue and remove the change to `watch_private_display`, I think this can be merged. As for resolving that one, I think it's as simple as declaring a `static void _rpn_display_string` that calls `watch_display_string` first, and then does your position 0 cleanup based on the first character in the buffer.

wryun:
Thanks - yup, I'll remove the change the watch_private_display. Sorry, meant to do that before after your earlier feedback.

One day I'll find someone else who thinks my binary number input method is genius rather than confusing!

joeycastillo (approved):

PR #114, opened 2022-10-30
Alarm face: make active alarm indicator show up only when there is an alarm set within the next 24 h

TheOnePerson:
I fell like spamming PRs for my alarm complication face, but after wearing it on the wrist for quite a while now, I found that two things needed some improvement.

1. As discussed with @wryun on #96, the active alarm indicator currently shows that there is _any_ alarm set, no matter the time or day. If you are working with alarms that fire off on specific days, this is not particularly useful. In these cases, it would be more helpful to see if there are any alarms set for an upcoming time period.
2. A minor UI issue: if you happen to (short) press the alarm button in normal mode, you are doomed to cycle through all 16 alarms to get back to where you were. A shortcut to jump back to the first alarm slot would be nice.

This PR fixes both issues:

On the simple clock watch face the alarm indicator is only shown, when there is an alarm active within the next 24 h set. (Give or take - this is only updated every 12 minutes to save battery power.)

If you hold the alarm button for more than 2 secs in normal mode, you are taken back to alarm slot no. 1. (Regular long press still toggles the enabled flag for an alarm.)

What else? I updated the comments to reflect the latest changes.

### DIFF
--- a/movement/watch_faces/complication/alarm_face.c
+++ b/movement/watch_faces/complication/alarm_face.c
@@ -37,8 +37,10 @@
 
     Usage:
     - In normal mode, the alarm button cycles through all 16 alarms. 
-    - Long pressing the alarm button in normal mode toggles the corresponding alarm on or off.
+    - Pressing the alarm button long in normal mode toggles the corresponding alarm on or off.
+      (Whereas pressing the alarm button extra long brings you back to alarm no. 1.)
     - Pressing the light button enters setting mode and cycles through the settings of each alarm.
+      (Long pressing the light button enters setting mode without illuminating the led.)
     - In setting mode an alarm slot is selected by pressing the alarm button when the slot number 
       in the upper right corner is blinking.
     - For each alarm slot, you can select the day. These are the day modes:
@@ -51,7 +53,8 @@
     - You can select the tone in which the alarm is played. (Three pitch levels available.)
     - You can select how many "beep rounds" are played for each alarm. 1 to 9 rounds, plus extra 
       long ('L') and extra short ('o') alarms.
-    - The simple watch face indicates any alarm set by showing the bell indicator.
+    - The simple watch face indicates if any alarm is set within the next 24h by showing the signal
+      indicator.
 */
 
 typedef enum {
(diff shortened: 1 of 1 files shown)


joeycastillo:
Looks good! I love the slight delay to show that the alarm wasn't in fact activated; it's a great subtle affordance to show the wearer what's going on.

wryun:
@TheOnePerson awesome. Next time I upgrade my watch, I'll be able to have more alarms (I set mine to 4 so it wasn't so painful to rotate through them), and I will enjoy the new 24h indicator behaviour!

movement/watch_faces/complication/alarm_face.c
```diff
@@ -268,7 +306,20 @@ bool alarm_face_loop(movement_event_t event, movement_settings_t *settings, void
             } else if (state->setting_state == alarm_setting_idx_minute) {
                         state->alarm[state->alarm_idx].minute = (state->alarm[state->alarm_idx].minute + 1) % 60;
             } else _abort_quick_ticks(state);
-        } else if (!state->is_setting) break; // no need to do anything when we are not in settings mode and no quick ticks are running
+        } else if (!state->is_setting) {
+            if (_wait_ticks >= 0 && _wait_ticks <= INT8_MAX) _wait_ticks++;
```
neutralinsomniac, line 310: _wait_ticks will always be <= INT8_MAX. I think you can remove it and just keep the first half of the conditional.

neutralinsomniac: unless you want to guarantee that _wait_ticks never goes negative here, which would obviously be solved by changing the `<=` to `<`

TheOnePerson: You are right, this is ought to be a `<` instead of `<=`. Thanks for finding this. This bug will not cause any trouble in real life, though. The erroneous line means that if you hold the alarm button for more than 255 seconds, the logic for jumping back to alarm slot no. 1 will be triggered again. But, as the wearer will already be sitting on alarm slot 1, nothing will happen. Nevertheless, I will fix this in one of the next PRs.

PR #115, opened 2022-10-31
Make countdown face use movement_play_alarm() instead of movement_play_signal()

neutralinsomniac:


joeycastillo:
Seems pretty straightforward; went ahead and tested on hardware just in case, a definite improvement. 

Thanks!

PR #116, opened 2022-11-01
Update README.md

eventi:
If you run a build with `make` the `emmake make` step will fail - There will probably be a more elegant solution in the makefile, but this works too

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ You may want to test out changes in the emulator first. To do this, you'll need 
 
 ```
 cd movement/make
+make clean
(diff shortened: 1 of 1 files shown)


wryun:
:+1: 

I've been meaning to make a PR to fix this properly (i.e. to have a separate build folder for the emulator (?)). Let me see if I can do that.

wryun:
See https://github.com/joeycastillo/Sensor-Watch/pull/118 (maybe).

eventi:
Closing in favor of #118

PR #118, opened 2022-11-01
Separate emscripten build directory

wryun:
This avoids the need to 'make clean' before you do a simulator build (or vice-versa).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,7 +44,7 @@ jobs:
         run: emmake make
         working-directory: 'movement/make'
       - name: Archive simulator build
-        working-directory: 'movement/make/build'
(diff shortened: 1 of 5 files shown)


eventi:
I tried this branch and had the same problem - it only occurs after running make

tahnok (approved):

.github/workflows/build.yml
```diff
@@ -44,12 +44,12 @@ jobs:
         run: emmake make
         working-directory: 'movement/make'
       - name: Archive simulator build
-        working-directory: 'movement/make/build'
+        working-directory: 'movement/make/build-sim'
```
eventi, line 47: Probably want this in the .gitignore

wryun: Yup, thanks!

PR #119, opened 2022-11-02
Add hours support to the countdown face

neutralinsomniac:


tahnok (commented):
I'm open to merging this as is if there's enough demand, but I feel fixing the deep sleep problem is a better solution

Maybe we disable the seconds when you enter hours? UX for that is kinda weird... We could silently ignore seconds but that also kind of sucks

PR #122, opened 2022-11-05
Play sound sequences asynchronously

TheOnePerson:
Hi folks, thanks to the help of @joeycastillo, I can now finalize this PR, which adds the functionality of playing sound sequences asynchronously and without blocking the ui. (A feature I need for another watch face I am currently working on.)

The idea is pretty simple: the function `void watch_buzzer_play_sequence(int8_t *note_sequence, void (*callback_on_end)(void))` takes a pointer to the sounds to be played and an optional pointer to a callback function, which will be invoked when the sequence has finished playing. A sound sequence consists of a series of tuples representing a) the buzzer note to be played plus b) the length, given in ~~32~~ 64 Hz ticks (1 tick is roughly ~~31~~ 16 ms). A trailing zero indicates the end of the sequence. Like so:
`static const int8_t _sound_seq_1[] = {BUZZER_NOTE_C7, 16, BUZZER_NOTE_E7, 16, BUZZER_NOTE_G7, 16, BUZZER_NOTE_C8, 24, 0};`

To play this sequence, you can do a simple `watch_buzzer_play_sequence((int8_t *)_sound_seq_1, NULL)`.

There is also a simple run-length-encoding logic implemented: if you present a negative number instead of a buzzer note, the sequence is re-wound by the given number of notes. The byte following a negative number determines the number of loops. 
I.e. if you want to repeat the last three notes of the sequence one time, you should provide the tuple `-3, 1`. The repeated notes **must not** contain any other repeat markers, or you will end up with an eternal loop.

Code sample how to use this:
```
#include "watch_buzzer.h"

static const int8_t _sound_seq[] = {BUZZER_NOTE_D5, 30, BUZZER_NOTE_REST, 2, -2, 4, BUZZER_NOTE_C7, 16, 0};

void play_sound(void) {
    watch_buzzer_play_sequence((int8_t *)_sound_seq, NULL);
}
```

It's not like playing midi sound files, but hey - it's a start! ;-)

~~**Edit:** This PR also implements a change to the logic of watch_rtc. You can use two new functions:
`int8_t watch_rtc_register_periodic_callback_slot(ext_irq_cb_t callback, uint8_t frequency);` let you register a callback function for a certain frequency using one of three freely available slots. The slot number (1 - 3) will be returned. If no free slot was available, the function returns -1.
Use `void watch_rtc_disable_periodic_callback_slot(uint8_t frequency, uint8_t slot);` to de-register a callback function in a callback slot.~~

~~**Edit no. 2:** The PR is currently being refactored to avoid the aforementioned changes to watch_rtc. It seems that the added logic impacts battery life noticeable, so I plan to use TC3 interrupt callbacks instead.~~

**Edit no. 3:** Refactoring is done now. The changes to the logic within watch_rtc are reverted, now TC3 is used for timing the callback for the buzzer sequences.


joeycastillo:
This sounds right. The issue isn’t that the interrupt stops the buzzer, it’s that the TCC, which controls the buzzer, is configured to not run in standby mode — so you only get a brief clip of tone every time the interrupt fires. 

Setting the RUNSTDBY bit in the TCC peripheral when the sequence starts, and un-setting it at the end should make this work.

TheOnePerson:
This sounds promising, thanks @joeycastillo! Will try this out.

TheOnePerson:
Short update: setting the RUNSTDBY bit via `hri_tcc_set_CTRLA_RUNSTDBY_bit(TCC0)` does not seem to do the trick. In fact, if I query the bit by `hri_tcc_get_CTRLA_RUNSTDBY_bit(TCC0)` it is set already. (And doing it the other way round - clearing at the beginning of the sequence and setting it back afterwards - does not have any effect either.) 
Am I using the wrong calls here?!

joeycastillo:
We also may need to set GCLK0 to run ONDEMAND for this to work. I can’t look at it right now but it should be a setting buried in the configuration files.

TheOnePerson:
I got it working (at least in principle), yay! Your hint to tcc RUNSTDBY did in fact do the trick. I just misread my debug output and did not realize that calling `hri_tcc_set_CTRLA_RUNSTDBY_bit(TCC0)` at the beginning of a sequence has not been successful. Now I can start working around it. Thanks so much!

TheOnePerson:
I just updated the description, the PR is ready now.

joeycastillo:
One question, sorry I didn't think to ask it at the same time yesterday: is there a reason to use the 32 Hz tick here instead of the 128 Hz tick? The reason I ask: if you use the 128 Hz tick you can avoid making any changes to Movement, and if you wanted to eventually add this capability to Movement (which admittedly is something I'd be wary of in the near term) you could use its internal fast tick to time this.

As it stands, it seems like this change to Movement would introduce a bug where if a watch face requests the 32 Hz tick for any reason, that tick would remain enabled even after the watch face resigns.

TheOnePerson:
My assumption was that the 128 Hz tick would interfere with the fast tick callback used by the movement. Am I mistaken that there can only be one callback function per frequency? My reasoning was that I wanted to tamper as little as possible with movement.c. Plus, I felt like choosing a tick frequency as low as possible, but still as fast as needed would help in regards to energy consumption.

And you are right: if a watch face requests a 32 Hz callback, it would not be disabled any more. The last commit "fixes" this - by not letting watch faces register 32 Hz ticks in the first place. It's not elegant, but integrating the buzzer sequence logic into movement.c seems to be a rather big change, with lots of opportunities for errors ;-)

joeycastillo:
Ah you’re quite right, callbacks, there can be only one. Would you be willing to refactor to use the 64 Hz tick instead? Reason being, it feels a little odd that 32 Hz ends up being this odd gap in the allowable tick rates for Watch faces, whereas saying the two fastest rates are claimed feels more reasonable.

TheOnePerson:
Of course! Having an arbitrary gap in the allowable tick rates feels somehow bumpy for me, too. As an alternative, I was thinking about extending the callback logic in watch_rtc.c. Something like so: have up to 4 callback functions per frequency ("slots"). One could register a callback by a function like this: `uint8_t watch_rtc_register_periodic_callback_slot(ext_irq_cb_t callback, uint8_t frequency)`. It returns the slot that has been assigned to the callback - or some indicator if no free slot was available. De-registering would look something like this: `void watch_rtc_disable_periodic_callback_slot(uint8_t frequency, unit8_t slot)`. I would not touch the existing logic built around `ext_irq_cb_t tick_callbacks[8]`, but rather use a new array `ext_irq_cb_t tick_callbacks_slots[8][4]`, combined with an `uint8_t tick_callbacks_slots_in_use[8]`, which would be an array of bitmasks to keep track of used or free slots for callbacks. The new logic for invoking these callback functions would have to be inserted into the `RTC_Handler()` routine. This, of course would be a somehow bigger refactoring and I am not sure if this approach would be far too complicated for what we will really ever make use of in the movement firmware.

What do think? Refactor to 64 Hz? Or extending the callback logic (and thus working on crucial parts of the firmware)?!

TheOnePerson:
The more I look into the code, the more feasible the approach with using several callback slots appears to me. I will try this out and then we can still decide, if this makes it into the main branch or has just been a test. Either way is fine by me.

TheOnePerson:
I just pushed the latest changes. The buzzer sequences do not collide with the tick callbacks used by the movement firmware anymore, due to some extensions to watch_rtc. I checked this on my watch and in the simulator and everything looks fine so far.

TheOnePerson:
I just implemented some changes to make this PR less invasive for the core logic located in watch_rtc. Now TC3 is used for timing the callback for the buzzer sequences.

TheOnePerson:
One last commit to this PR: there was an error in the timer frequency, leading to sequences being played too slowly.

joeycastillo:
Sorry for the delay in reviewing this! I wanted to have a chance to test it myself and run some power consumption numbers as well, but the good news is it seems all good! 

Going to merge this in first and then merge in the interval face; thanks so much for these PR's, it's quite cool what you've done here! 😎

PR #126, opened 2022-11-08
append beep sound to counter_face

ShogoOkamoto:
I created this 'counter_face' to keep track of the number of laps in a jog. After installing it on an actual watch and using it, I found it difficult to use, so I improved it to beep sound based on the number of laps. It seems to be especially effective when running at night.


joeycastillo:
Looks good to me! 👍

PR #130, opened 2022-11-13
Interval timers face

TheOnePerson:
This face brings 9 customizable interval timers to the sensor watch, to be used as hiit training device and/or for time management techniques. It relies on PR #122 to work (all chimes are played asynchronously).

- There are 9 interval timer slots, you can cycle through these with the alarm button (short press). For each timer slot, a short "slideshow" displaying the relevant details (like length of each phase - see below) is shown.

- To start an interval timer, press and hold the alarm button.

- To pause a running timer, press the alarm button (short press).

- To completely abort a running timer, press and hold the alarm button.

- Press and hold the light button to enter settings mode for each interval timer slot.

- Each interval timer has 1 to 4 phases of customizable length like so:
  (1) prepare/warm up --> (2) work --> (3) break --> (4) cool down.
  When setting up or running a timer, each of these phases is displayed by the letters "PR" (prepare), "WO" (work), "BR" (break), "CD" (cool down).

- Each of these phases is optional, you can set the corresponding minutes and seconds to zero. But at least one phase needs to be set, if you want to use the timer.

- You can define the number of rounds either only for the work phase and/or for the combination of work + break phase. Let's say you want an interval timer that counts 3 rounds of 30 seconds work, followed by 20 seconds rest:
		work 30s --> work 30s --> work 30s --> break 20s
  You can do this by setting 30s for the "WO"rk phase and setting a 3 in the lower right hand corner of the work page. The "LAP" indicator lights up at this position, to explain that we are setting laps here. After that, set the "BR"eak phase to 20s and leave the rest as it is.

- If you want to set up a certain number of "full rounds", consisting of work phase(s) plus breaks, you can do so at the "BR"eak page. The number in the lower right hand corner determines the number of full rounds to be counted. A "-" means, that there is no limit and the timer keeps alternating between work and break phases.

- This watch face comes with several pre-defined interval timers, suitable for hiit training (timer slots 1 to 4) as well as doing
  work according to the pomodoro principle (timer slots 5 to 6). Feel free to adjust the timer slots to your own needs (or completely wipe them ;-)


TheOnePerson:
Updated underlying buzzer sequences logic.

PR #131, opened 2022-11-15
Add quick counting through hours, minutes, etc. to set time face.

TheOnePerson:
As the title says: I modified the set time face, so that you can now quick cycle through various settings by holding the alarm button. (And also setting the date to February 29th on a leap year is possible now.)


joeycastillo:
Thanks for this! I'm seeing an issue when testing on hardware: I can get into a state after holding the alarm button where the unit keeps cycling after releasing the button, it just runs away until I press the Alarm button again. It looks like this may not be an error in the settings watch face though: in testing it appears that in some cases neither `EVENT_ALARM_LONG_UP` or `EVENT_ALARM_BUTTON_UP` are getting passed along to the watch face? 

Need to test more but let me know if you can replicate this; it involves holding the Alarm button to cycle through the hours, and releasing it around the one second mark.

TheOnePerson:
Yes, I can reproduce this as well. It indeed seems like sometimes movement does neither fire the `_LONG_UP` nor the `_BUTTON_UP` event when the alarm button is released. I am not sure how to fix this, since I cannot find a logic or flow problem in the movement code. One could maybe build in a fail safe like checking the button's pin level in the settings watch code, but this would certainly be the opposite of elegant... :-/

TheOnePerson:
OK, I just added checking the pin level as a failsafe for hanging quick cycling. This looks fine now on my hardware.

movement/watch_faces/settings/set_time_face.c
```diff
@@ -97,7 +97,9 @@ bool set_time_face_loop(movement_event_t event, movement_settings_t *settings, v
 
     switch (event.event_type) {
         case EVENT_TICK:
-            if (_quick_ticks_running) _handle_alarm_button(settings, date_time, current_page);
+            if (_quick_ticks_running) 
+                if (watch_get_pin_level(BTN_ALARM)) _handle_alarm_button(settings, date_time, current_page);
+                else _abort_quick_ticks();
```
joeycastillo, line 102: Appreciate this stopgap! I'll test some more this weekend in the hopes of merging it in; I'd love to fix the underlying issue, but after a couple of hours of chasing it today I'm pretty flummoxed, and as workarounds go, this is pretty minimal :)

My only thought would be to wrap the inner if/else in braces; the indent feels clear to me, but if someone comes in to add something later, it could lead to confusion or glitchiness (cf. [goto fail](https://dwheeler.com/essays/apple-goto-fail.html))

TheOnePerson: You are right regarding the missing braces. I saw this yesterday, too, and have fixed it in the last commit of this PR.

movement/watch_faces/settings/set_time_face.c
```diff
@@ -72,22 +52,79 @@ bool set_time_face_loop(movement_event_t event, movement_settings_t *settings, v
...
-                    date_time.unit.day = date_time.unit.day + 1;
-                    // can't set to the 29th on a leap year. if it's february 29, set to 11:59 on the 28th.
-                    // and it should roll over.
-                    if (date_time.unit.day > days_in_month[date_time.unit.month - 1]) {
-                        date_time.unit.day = 1;
+        case 5: { // day
+            uint32_t tmp_day = date_time.unit.day;   // use a temporary variable to avoid messing up the months
+            tmp_day = tmp_day + 1;
+            // handle February 29th on a leap year
+            if (((tmp_day > days_in_month[date_time.unit.month - 1]) && (date_time.unit.month != 2 || (date_time.unit.year % 4) != 0))
```
tahnok, line 59: I think there's an `is_leap` in watch_utility that will leap year calculation fully

movement/watch_faces/settings/set_time_face.c
```diff
@@ -29,32 +29,12 @@
 #define SET_TIME_FACE_NUM_SETTINGS (7)
 const char set_time_face_titles[SET_TIME_FACE_NUM_SETTINGS][3] = {"HR", "M1", "SE", "YR", "MO", "DA", "ZO"};
 
-void set_time_face_setup(movement_settings_t *settings, uint8_t watch_face_index, void ** context_ptr) {
-    (void) settings;
-    (void) watch_face_index;
-    if (*context_ptr == NULL) *context_ptr = malloc(sizeof(uint8_t));
-}
+static bool _quick_ticks_running;
```
tahnok, line 32: IMO this variable would be better in a state struct that is stored in the *context ptr like other faces. Would have to migrate the current use of the context ptr for the page number to it, but that seems straightforward

tahnok (commented):
Got a bit nitpicky here, sorry but I love this functionality and am looking forward to getting it merged

movement/watch_faces/settings/set_time_face.c
```diff
@@ -29,32 +29,12 @@
...
-void set_time_face_activate(movement_settings_t *settings, void *context) {
-    (void) settings;
-    *((uint8_t *)context) = 0;
-    movement_request_tick_frequency(4);
-}
-
-bool set_time_face_loop(movement_event_t event, movement_settings_t *settings, void *context) {
-    uint8_t current_page = *((uint8_t *)context);
+static void _handle_alarm_button(movement_settings_t *settings, watch_date_time date_time, uint8_t current_page) {
+    // handles short of long pressing of the alarm button
```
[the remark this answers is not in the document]
TheOnePerson: Of course, will fix.

movement/watch_faces/settings/set_time_face.c
```diff
@@ -72,22 +52,79 @@ bool set_time_face_loop(movement_event_t event, movement_settings_t *settings, v
                 case 4: // month
                     date_time.unit.month = (date_time.unit.month % 12) + 1;
                     break;
-                case 5: // day
-                    date_time.unit.day = date_time.unit.day + 1;
-                    // can't set to the 29th on a leap year. if it's february 29, set to 11:59 on the 28th.
-                    // and it should roll over.
-                    if (date_time.unit.day > days_in_month[date_time.unit.month - 1]) {
-                        date_time.unit.day = 1;
+        case 5: { // day
```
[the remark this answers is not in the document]
TheOnePerson: You are right, indentation is off (above and below the case-5 part). Will fix. (As for the braces: I got strange compiler errors without these...)

movement/watch_faces/settings/set_time_face.c
TheOnePerson: I looked and could not find such a function. Besides, the condition for checking for leap years is rather trivial, so I am not sure using a function would be necessary here.

movement/watch_faces/settings/set_time_face.c
TheOnePerson: My approach was to only store variables that need some persistence to the state struct. The quick ticks running indicator is only needed when the settings face is the active watch face and is no longer needed when resuming it. So just using a static var seems to be the more straightforward strategy to me.

PR #133, opened 2022-11-15
Set bell indicator on reactivating countdown/tomato faces

wryun:
If you moved away and back to these faces, you'd lose the indicator. Not that the indicator here seems very valuable, since it's not displayed on simple_clock_face, but at least this makes things more consistent.


wryun:
Cleaner to just put this in the draw functions, but this seemed like the less invasive bugfix.

PR #134, opened 2022-11-15
Add Sensor board with BMI270 IMU

abhignay:
As the title of this PR suggests, I've created a sensor board using the BMI270 IMU, it should be somewhat better than the LIS2DW as it includes a tri-axis gyroscope and a tri-axis accelerometer!

Silkscreen is rudimentary at this point, I'm working on fixing that and the schematic documentation!


joeycastillo:
Awesome, looks cool! 😃

PR #135, opened 2022-11-16
vendor external libs

bademux:
Can we please discuss dropping git modules in favor of direct vendoring external depenencies.

**Why:**
git-modules system is hard to handle (imho) both for managing dependencies and automation.
For automation it introduces additional dependency - git.
Git modules exclude us from nice github\gitlag features like direct snapshot download:
```bash
https://github.com/joeycastillo/Sensor-Watch/tarball/main/archive/main.tar.gz 
```
We can't use Dockerfile ADD command to download sources.

Updating dependencies can be less strict with vendoring way, please see [README.md](https://github.com/joeycastillo/Sensor-Watch/compare/main...bademux:Sensor-Watch:feature/vendoring_ext_libs?expand=1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R79) as proposed way of updating deps

littlefs [tag v2.5.0](https://github.com/littlefs-project/littlefs/tree/40dba4a556e0d81dfbe64301a6aa4e18ceca896c) is used for vendoring
 tinyusb [commit f8288be](https://github.com/hathach/tinyusb/tree/f8288be03f28ad7b944e6925f49422dfa39202c5) is used for vendoring; only hw/mcu/microchip submodule


tahnok:
I'm not a fan of this. I agree that git submodules are pain for users and for automations, but I think they more clearly delineate the boundary between SensorWatch code and external stuff and it's pretty easy to update the external deps when needed

joeycastillo:
For the moment I'd like to hold off on this; I see the case for both ways of going about it, but for now the submodule method is working well enough, and it's not proving to be a significant barrier to entry for folks who are trying to make contributions. I'm hesitant to make a big change to how the project is structured and built — and potentially the documentation surrounding that — when there's a lot of work that needs to go into the firmware itself right now.

Let's revisit this at a later date.

PR #140, opened 2022-11-20
Add stock stopwatch face

TheOnePerson:
This PR brings the original F-91W stopwatch functionality including counting hundredths of seconds and lap timing.

There are two improvements compared to the original stopwatch functionality:

1. When reaching 59:59 the counter does not simply jump back to zero, but keeps track of hours in the upper right hand corner. (Up to 24h)
2. Long pressing the light button toggles the led behavior: it either turns on on each button press or it doesn't.

Some technical remarks:

- This watch face needs PR #122 to work. It registers a 128 Hz callback for counting the fractions of a second, which would not be possible by calling `movement_request_tick_frequency()`.
- As stated above: a running stopwatch means that we have a constant, although minimalistic callback running at 128 Hz. I do not  have the equipment to measure it exactly, but I would assume that this put a certain load on the battery.


wryun:
It would be nice to know how much power this ends up drawing. One way we could get at it an estimate - perhaps - is to try to measure the average idle time between ticks.

movement/watch_faces/complication/stock_stopwatch_face.c
```diff
@@ -0,0 +1,246 @@
...
+static void _display_ticks(uint32_t ticks) {
+    // show minutes, seconds and fractions derived from 128 Hz tick counter
+    char buf[14];
+    uint8_t sec_100 = (ticks & 0x7F) * 100 / 128;
+    uint32_t seconds = ticks >> 7;
+    uint32_t minutes = seconds / 60;
+    if (_hours)
+        sprintf(buf, "%2u%02lu%02lu%02u", _hours, minutes, (seconds % 60), sec_100);
+    else
+        sprintf(buf, "  %02lu%02lu%02u", minutes, (seconds % 60), sec_100);
```
wryun, line 73: The clock face does a bit of work to figure out what updates are necessary; this is likely to pay off more here, I suspect.

movement/watch_faces/complication/stock_stopwatch_face.c
```diff
@@ -0,0 +1,248 @@
...
+static uint32_t _lap_ticks;
+static uint8_t _blink_ticks;
+static uint32_t _old_seconds;
+static uint8_t _old_minutes;
+static uint8_t _hours;
+static bool _colon;
+static bool _is_running;
+
+static void _fast_tick_callback() {
+    _ticks++;
```
wryun, line 53: I was thinking about this, and wondering about 'missed' ticks (which AFAIK will happen now if anything takes more than 1/128th of a second inside a tick). Since the RTC only gives us seconds (I think?), the only way I could think of to handle this is to count the tick offset till the first RTC time change - i.e. read the RTC time in the fast_tick_callback - and on every subsequent RTC time change reset the ticks as appropriate. This means we are out by at most a couple of ticks that occur in the first fraction of a second, and are unlikely to see an interruption. We could also use this to allow entering sleep mode while the stopwatch was on or at least removing the callback while the face wasn't active.

Thoughts? Maybe this is unnecessary complexity (or I'm simply not understanding how ticks work).

wryun: Hmm, looks like we can actually use one of the timer counters, and simply read out the current value when we want to update the display. This seems like a more straightforward way of figuring out the time.

It looks (maybe) like TC1 is unused, based on a quick look at the code. You can see how TC0 is configured to do 1000hz callbacks in watch_private.c.

TheOnePerson: Your idea of using TC1 for counting sub-second intervals and simply readout values instead of using movement callbacks seems intriguing! My problem is that do not really have an understanding of how this stuff works... ;-) 
But I will take a deeper look and maybe I can figure out how to set up the counter unit and read out values properly, and if it works, this appears to be the cleaner approach. I still don't know if it was a better approach battery-wise, though. 
Will set this PR to draft in the meantime.

wryun: I'm almost certainly more clueless than you, and have no real experience with embedded stuff; I seem to remember running sleeps meant none of the rtc ticks were executed during the period I was running the sleep, but I may have misinterpreted what was going on.

I also know nothing about the TC stuff other than what I've just been trying to read in the code and the documentaion. For example, see: https://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-42123-SAM-Timer-Counter-(TC)-Driver_ApplicationNote_AT03263.pdf (in particular, tc_get_count_value)

Note that this refers to a function in the older SDK, and unfortunately there doesn't seem to be any helpers in the current one; i.e. you have to read/write to the appropriate places manually, as in watch_private.c. However, if you download the v3 sdk you can see what those library functions do under the hood, and there's also example code.

wryun: `hri_tccount32_get_COUNT_COUNT_bf` looks like it might be useful.

TheOnePerson:
So, after fiddling around for days with the option of using the TC peripherals to do the time counting, I decided to stick with the current way of implementation: the 128 Hz timer callback. I added some extra logic to handle waking up a running timer from sleep mode and call this PR ready for now.

wryun:
My apologies :( Did it simply not work, or did it make the code a mess, or ... ?

TheOnePerson:
No need to apologize for anything! I tried several ways using TC2 or TC1, but in the end, I could not solve the problem of stopped timing in low energy mode. 

The following lines of code enable the TC1 timer:
```
    // clock TC1 with the 4 MHz clock on GCLK0.
    hri_gclk_write_PCHCTRL_reg(GCLK, TC1_GCLK_ID, GCLK_PCHCTRL_GEN_GCLK0_Val | GCLK_PCHCTRL_CHEN);
    // and enable the peripheral clock.
    hri_mclk_set_APBCMASK_TC1_bit(MCLK);
    // disable and reset TC1.
    hri_tc_clear_CTRLA_ENABLE_bit(TC1);
    hri_tc_write_CTRLA_reg(TC1, TC_CTRLA_SWRST);
    // configure the TC to 15.625 KHz
    hri_tc_write_CTRLA_reg(TC1, TC_CTRLA_PRESCALER_DIV64 |
                                TC_CTRLA_MODE_COUNT32 | 
                                TC_CTRLA_PRESCSYNC_PRESC |
                                TC_CTRLA_RUNSTDBY);
```

Just starting the timer and reading count values using ``hri_tccount32_read_COUNT_reg(TC1)`` somehow did not work. The count values did not seem to start at 0 and they did not increment as expected, and I was not able to fix that (e.g. by using ``hri_tccount16_write_COUNT_reg(TC1, 0)`` to reset the values.

So the second approach was to enable a periodic interrupt, based on the TC1 counting. This did work in general - but when the watch entered low energy mode, the interrupt is not executed anymore. So, at the end, this method did not make things easier or more elegant than the current implementation with a "regular" 128 Hz callback.

So now the stop watch counts the 128 Hz ticks offset compared to RTC seconds to figure out how long it has been sleeping. This adds some complexity to the 128 Hz callback, but unfortunately, I was not able to solve this problem more elegantly.

If somebody has a good idea leading to a more straightforward approach, I would be more than glad! :-)

TheOnePerson:
> For now I can just make it so TC2 and TC3 aren't shut down in low energy mode; they currently aren't turned on at all, so there'll be no power impact for watch faces that don't use them.

That would be perfect! I do understand your concerns regarding the additional load of cycling through four possible callbacks for each of the eight prescaler interrupts. In fact, I am using this version of the firmware on my own watch and I had to change the battery just recently. I am afraid there is a real live impact to battery life...

Anyway: Having TC2 & TC3 as two usable timers that keep on running even in low enery mode would be useful for a lot of use cases, I guess. I will set this PR to draft for now, refactor the whole thing and come back when it's done.

Besides, I will also refactor pr #122 so that it uses TC3 for the 64 Hz callback and thus get rid of the changes in `watch_rtc.c`

joeycastillo:
Hey so on further review, it looks like I'm not shutting down the TC peripherals. I thought I was (because there's a function called `_watch_disable_all_peripherals_except_slcd`) but it currently only hits the TCC, ADC, EIC and SERCOMs. Which means I think you should be able to set up TC3 to run in low energy mode without any changes: 

```c
MCLK->APBCMASK.reg |= MCLK_APBCMASK_TC3;
GCLK->PCHCTRL[TC3_GCLK_ID].reg = GCLK_PCHCTRL_GEN_GCLK3 | GCLK_PCHCTRL_CHEN;
TC3->COUNT16.CTRLA.reg |= TC_CTRLA_RUNSTDBY;
```

GCLK3 is hooked up to the quartz crystal, and it's configured to run on demand so if any peripheral is requesting it (which that second line does), GCLK3 should keep clocking the TC even in low energy mode if RUNSTDBY is set. Let me know if this works!

joeycastillo:
It's entirely possible we should write some functions to set this sort of thing up deeper in the watch library; there are several ways to use the TC peripherals (timers like this in addition to signal generation and frequency capture) and it would be nice to provide a higher level API. Good TODO item for the future!

TheOnePerson:
I finally managed to refactor the code, so that it uses TC2 as a source for the 128 Hz callback.

PR #141, opened 2022-11-22
Add support for bicolor led in the simulator

TomHodson:
I noticed that the bicolor led isn't supported in the simulator. Here is a quick fix for that. 

I added an SVG filter to the #light rectangle in the SVG which lets us multiply the colours by a matrix in code. I then insert values so that the original green colour gets mixed into the combination of red and green requested by `watch_set_led_color`

### Testing
So far I've only tested this with the standard firmware emscripten build where it seems to work. As far as I can tell, this doesn't touch any code that would run on real hardware.

### Possible issues
I wasn't 100% sure what to do when red + green > 255, current behaviour is that it gets clamped to 255 in the simulator. I'm also changing both the opacity and the colours at the same time. This is almost certainly not the physically correct way to do this but it looks reasonable!

### DIFF
--- a/watch-library/simulator/shell.html
+++ b/watch-library/simulator/shell.html
@@ -51,6 +51,14 @@
         <stop offset="0.94" stop-color="#000d00" stop-opacity="0.05"/>
         <stop offset="1" stop-opacity="0"/>
       </radialGradient>
+      <filter id="ledcolor">
+        <feColorMatrix in="SourceGraphic" type="matrix"
+                       values=" 0      0      0      0 0
+                                0      1      0      0 0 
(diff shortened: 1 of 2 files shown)


joeycastillo:
This looks great! Thanks so much for doing this, I was never sure how to make it work :)

dglaude:
Don't we have a limited set of watch that have different led color?

TomHodson:
We could fix that by adding a check for the environment variable that specifies it's the special red/blue watch and adding a line like
```javascript
color_matrix[11].value = $1 / 255; // blue value
```
with extra logic to decide whether to light the green or blue channel.

PR #144, opened 2022-11-24
Adds simulator watch skin toggle for model A158WEA-9

kyleboyle:
Should one be feeling classy...
This isn't a perfect replication as the button font doesn't match exactly and the lithium label isn't there because the font is completely different (I couldn't reuse existing shapes).

[Kapture 2022-11-24 at 18.50.49.webm](https://user-images.githubusercontent.com/6877108/203871838-9e82a0eb-b57b-409e-b9de-fa5fc151b488.webm)


joeycastillo:
This is so rad! And the A158WEA-9 is my daily driver, so I especially love that this is the one you've added :)

Thanks for the PR!

PR #149, opened 2022-12-02
Week number watch face

slim:
This watch face is an alternate simple clock, it's exactly the same, but displays iso 8601 week number instead of seconds count.

N.B. did not test this on real hardware, my sensor-watch is hopefully coming in the next batch :)


joeycastillo:
This looks good! Simple, gets the job done. Thanks for the PR!

joeycastillo:
ALSO! Thanks for the new utility functions. These are going to be really helpful; I know that space folks are SUPER into day-of-year stuff, and the days since new year function is going to be super useful! 😃

PR #152, opened 2022-12-22
Precision watch update

BarsMonster:
* nanosec face - applies static or temperature-compensated correction. 
* finetune face - allows to adjust watch with sub-second precision, and calculate/apply static ppm frequency offset into nanosec
* set_time_hackwatch - sets seconds (almost) precisely at 0 subsecond counter. Quality of life improvements - long presses to adjust in other direction. Start adjustment from year, so that seconds are adjusted last. 
* databank - allows to show static data for your reference. 
* tempchart - stores temperature statistics across hours/temperature to allow improvements in watch precision in future generations. 

Few changes in common libs:
* watch_rtc - new methods for freqcorr and start/stop RTC reliably
* watch_display_character_lp_seconds - hopefully slightly improves power efficiency of simple_clock without power saving.
* No longer switching off segments unnecessarily
* All 8 periodic events are checked. Debatable if it is needed. 
* Long press do not return to face 0 - as I really need this for lots of adjustments in my faces


BarsMonster:
Please note that I was also changing sunrise_sunset_face.c to use .ini file for location in the middle of the work, but reverted back to current release version before PR. You might still see it in the history. 

This needs more work to make sure all location-aware faces get data from ini file.

BarsMonster:
@neutralinsomniac 
1) databank and tempchart can definitely be separate. So we can easily only merge in everything except it if there are any concerns with these. 
2) My concern against making it default is user experience is different vs original Casio. Improved precision will not help much without precision time infrastructure, as ~1-1.5 seconds of error per day will quickly destroy any benefits.

joeycastillo (approved):
This looks good! I will confess I haven't had time to fully grok the usage, but seeing the significant progress folks are making in the Discord thread, I want to get this merged in now and play with it in main. Feel free to open another PR if you have additional changes!

Also: I know that this disables the long press of MODE to return to the first watch face. @BarsMonster and I have discussed a strategy for restoring it without affecting the ability for faces like his to override it as needed; we'll do that in a future commit.

PR #165, opened 2023-01-09
Add Tachymeter face

rcassani:
The Tachymeter complication emulates the [tachymeter](https://en.wikipedia.org/wiki/Tachymeter_(watch)) function often present in watches, that computes the average speed in [units per hour] for a given distance given in [units].

# Use case
1. User sets the distance
2. User starts the tachymeter when the trip begins
3. User stops the tachymeter when the trip ends
4. The watch presents the average speed and trip duration in seconds

## Use
1. Go to tachymeter face, `TC` is shown in the Weekday Digits 
2. A steady `d` in the Day Digits indicates the distance to be used. To edit the distance:
    1. Long-press the Alarm button, the distance edition page (`d` will blink)
    2. Use the Light button to change the editing (blinking) digit, and press Alarm to increase its value
    4. Once done, long-press the Alarm button to exit the distance edition page
3. Press the Alarm button to start the tachymeter. A running animation will appear in the Day Digits
4. Press the Alarm button to stop the tachymeter
6. The average speed and total time information will alternate. The average speed will be shown alongside `/h` in the Day Digits; and the total time will be shown alongside `t` in the Day Digits.
7. Long press the Light button to return to the distance `d` page, and restart the tachymeter from there.
    * Long-press the light button in the steady distance page to reset the distance to 1.00

## Pending design points

*. `movement_request_tick_frequency(4) is used to obtain a 4Hz ticking, thus having a time resolution of 250 ms. Not sure if using `event.subsecond` is the proper way to get the fractions of second for the start and final times.

*. For distance and average speed, the Second Digits (position 8 and 9) can be seen as decimals, thus possible to show distances as short as 0.01 km (or miles) and speeds as low as 0.01 km/h (or mph). However, if the same idea is used for the total time (showing hundredths), this limits the display to 9999.99 seconds (~2h:45m).

### DIFF
--- a/movement/make/Makefile
+++ b/movement/make/Makefile
@@ -74,6 +74,7 @@ SRCS += \
   ../watch_faces/demo/frequency_correction_face.c \
   ../watch_faces/complication/alarm_face.c \
   ../watch_faces/complication/ratemeter_face.c \
+  ../watch_faces/complication/tachymeter_face.c \
 # New watch faces go above this line.
 
 # Leave this line at the bottom of the file; it has all the targets for making your project.
--- a/movement/movement_faces.h
+++ b/movement/movement_faces.h
@@ -59,6 +59,7 @@
 #include "alarm_face.h"
 #include "ratemeter_face.h"
 #include "weeknumber_clock_face.h"
+#include "tachymeter_face.h"
 // New includes go above this line.
 
 #endif // MOVEMENT_FACES_H_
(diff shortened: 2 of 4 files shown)


joeycastillo (approved):
Nice! I just made a couple of changes to silence compiler warnings; the only other change we might want to make is to have a long press of the Alarm button clear the tachymeter, instead of a long press on Light; it feels like this would have more consistency with other watch faces (long press Alarm to clear, long press again to change the distance). But there's no need for that to block merging.

Thanks for making this!

PR #169, opened 2023-01-11
Implement tarot card face

neutralinsomniac:


strickinato:
😮 -- love this!

joeycastillo:
This is so cool! I made a couple of changes for legibility; let me know what you think. Mostly it's to make "m" characters where I could; I think they look legible except maybe in “Emprs” and “Empor”. Also tweaked “Draw” and “World” to be a little more W-ish. 

Open to your feedback though as the designer of the watch face!

PR #176, opened 2023-01-13
alarm: fix accidental assignment vs conditional logic

neutralinsomniac:


TheOnePerson:
Oh, this was originally my bad. Thanks for fixing it!

PR #177, opened 2023-01-13
Minor enhancements for the countdown face

rieck:
First of all, I like to thank you all for this project. It is so much fun working with the watch and the firmware. While playing with the different faces, I noticed that the countdown face behaved slightly differently than expected. I think it should work similarly to the stopwatch and support _pausing_ as well as _resetting_ to the entered countdown value. Therefore, I wrote a few minor enhancements:

- The countdown can be paused using the alarm button, similar to the stopwatch face.

- The state of the countdown face is indicated in the top right: `P` = pause, `S` = setting, ` ` = running

- The last entered countdown can be restored by long pressing the light button in setting or waiting (pause) mode.

- The countdown can be reset to zero by long pressing the alarm button in setting or waiting (pause) mode.

- Once the countdown reaches 0, it automatically switches back to the last entered value and pauses.

I'm not sure if this is a useful extension, but I thought I'd prepare a pull request anyway ;-)

### DIFF
--- a/movement/watch_faces/complication/countdown_face.c
+++ b/movement/watch_faces/complication/countdown_face.c
@@ -1,6 +1,7 @@
 /*
  * MIT License
  *
+ * Copyright (c) 2023 Konrad Rieck
  * Copyright (c) 2022 Wesley Ellis
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -30,6 +31,24 @@
 #include "watch.h"
 #include "watch_utility.h"
 
+/*
+    Slight extension of the original countdown face by Wesley Ellis.
+
+    - The countdown can be paused using the alarm button, similar to
+      the stopwatch face.
+
+    - The state of the countdown is indicated in the top right: 
+      "P" = pause, "S" = setting, " " = running
+
+    - The last entered countdown can be restored by long pressing the
+      light button in setting or waiting (pause) mode.
+
+    - The countdown can be reset to zero by long pressing the
+      alarm button in setting or waiting (pause) mode.
+
+    - Once the countdown reaches 0, it automatically switches back
+      to the last entered value and pauses.
+*/
 
 #define CD_SELECTIONS 3
 #define DEFAULT_MINUTES 3
@@ -37,6 +56,26 @@
 
 static inline int32_t get_tz_offset(movement_settings_t *settings) {
     return movement_timezone_offsets[settings->bit.time_zone] * 60;
+}
+
+static inline void store_countdown(countdown_state_t *state) {
+    /* Store set countdown time */
+    state->set_hours = state->hours;
+    state->set_minutes = state->minutes;
+    state->set_seconds = state->seconds;
+}
+
+static inline void load_countdown(countdown_state_t *state) {
+    /* Load set countdown time */
+    state->hours = state->set_hours;
+    state->minutes = state->set_minutes;
+    state->seconds = state->set_seconds;
+}
+
+static inline void reset_countdown(countdown_state_t *state) {
+    state->hours = 0;
+    state->minutes = 0;
+    state->seconds = 0;
 }
 
 static void start(countdown_state_t *state, movement_settings_t *settings) {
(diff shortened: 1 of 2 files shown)


neutralinsomniac:
Hmm. I don't think I've ever used a countdown face that behaves like this. But I agree with you that the current implementation is also unfamiliar. From my experience with countdown faces on other watches, it usually works like this (assuming 2 button control):
button A enters settings mode and works exactly how it works now.
button B starts the timer if it's not running, or pauses/unpauses if it is running. while paused, button A resets the timer to the last set value. this is how the stopwatch face operates as well.

I don't think the state indicator is necessary either.

Basically I would love if this face mirrored the countdown timer behavior on any existing casio watch with the feature.

rieck:
Thanks for the feedback. I have never used a countdown timer on a Casio watch. Pressing button A when paused resets the timer to the last value and at the same time enters the settings mode. Is this correct?

neutralinsomniac:
Not at the same time. button A (in our case the light button) while the timer is paused will reset the timer only. then if you press button A again it enters settings mode. also if you press button A while the timer is running, nothing happens.

rieck:
Makes sense. Here is an update:

- Long button presses have been removed.

- The light button is used to restore the last countdown as well as enter setting mode when pressed again.

- The visual state has been removed from the display.

- The internal states have been renamed to reflect the new logic.

movement/watch_faces/complication/countdown_face.c
```diff
@@ -203,15 +245,6 @@ bool countdown_face_loop(movement_event_t event, movement_settings_t *settings,
         case EVENT_BACKGROUND_TASK:
             ring(state);
             break;
-        case EVENT_ALARM_LONG_PRESS:
```
neutralinsomniac, line 206: I know this doesn't mirror the original firmware, but I think we should keep this quick reset around. Otherwise it takes a very large number of button presses to input a timer that is less than what is currently showing. Or we implement hold to quickly advance the current setting, which is more in line with most timer behavior on other watches. If you wanted to try to tackle that, [the set time face](https://github.com/joeycastillo/Sensor-Watch/blob/main/movement/watch_faces/settings/set_time_face.c#L105) shows how that can be done. If that seems like too much to chew off, I'd just restore this block for now.

movement/watch_faces/complication/countdown_face.c
```diff
@@ -93,15 +119,24 @@ static void draw(countdown_state_t *state, uint8_t subsecond) {
...
     watch_clear_indicator(WATCH_INDICATOR_BELL);
+    load_countdown(state);
 }
 
 static void ring(countdown_state_t *state) {
     movement_play_alarm();
-    reset(state);
+    movement_cancel_background_task();
+    watch_clear_indicator(WATCH_INDICATOR_BELL);
+    state->mode = cd_paused;
```
neutralinsomniac, line 139: Can you also restore the saved timer value here? This is what other casio timers usually do too :)

neutralinsomniac:
Did a quick test in the simulator. Holding down alarm in order to quickly increase settings appears to increment the setting by two's instead of one's (presumably because it's ticking too fast). Cursory glance I don't see anything obviously different from the set time face (which seems to work fine). Are you able to get the simulator running to help experiment with your code?

rieck:
Good catch. The simulator shows the same problem. I could not spot an obvious issue, so likely it is some subtle logic flaw in switching the tick frequency. I'll check next weekend.

movement/watch_faces/complication/countdown_face.c
```diff
@@ -30,15 +31,58 @@
...
+    /* Load set countdown time */
+    state->hours = state->set_hours;
+    state->minutes = state->set_minutes;
+    state->seconds = state->set_seconds;
+}
+
+static inline void button_beep(movement_settings_t *settings) {
+    // play a beep as confirmation for a button press (if applicable)
+    if (settings->bit.button_should_sound)
+        watch_buzzer_play_note(BUZZER_NOTE_C7, 50);
```
neutralinsomniac, line 83: BUZZER_NOTE_C8 :)

neutralinsomniac:
Nice work. I think these enhancements bring this face to about as close as we can get to stock casio functionality. Hoping this gets merged :)

joeycastillo:
Tested this a bit today and it works great! Feels very intuitive :)

Thanks for opening this pull request, stoked to have it merged!

PR #178, opened 2023-01-14
Flashlight

fedpost:
It's pretty much as stated in my README, if you find even the slightest value or humor in having a flashlight program on an F-91W based watch, this will do that!


justinmbrock:
I would put this face on my watch, but I think it needs to be resubmitted without editing the README.md if it's going to get merged.

joeycastillo:
Sorry for the delay in responding to this PR. I'm hesitant about merging this in because it is substantially similar to the blinking light face, even keeping the same BL title and Fast/Slow indicator (even though this watch face doesn't blink the LED anymore).

My main issue, though, is that the green LED on the Sensor Watch board is unlikely to function usefully as a torch or flashlight, and I'm hesitant to merge in a watch face with substantial negative battery impact when the value isn't clear. Especially because folks who aren't aware that its intent is mostly humorous may add this watch face expecting it to be useful, and find it mostly drains their battery.

I'm sorry if this isn't the response you had hoped for, but I'm afraid I'm going to close this one.

PR #180, opened 2023-01-14
Restore long press on MODE to return to first watch face by adding a handler for unhandled events

joeycastillo:
This PR adds a new `movement_default_loop_handler` function that takes care of standard Movement behaviors. The intent behind this refactor is to have all watch faces call the default loop handler in the `default:` case of their `watch_face_loop` functions. 

At this time, the default handler handles three events: 

* `EVENT_MODE_BUTTON_UP` advances to the next watch face.
* `EVENT_LIGHT_BUTTON_DOWN` illuminates the LED.
* `EVENT_MODE_LONG_PRESS` returns to the first watch face.

This reduces the amount of boilerplate code in watch faces: by simply omitting cases for these events, the switch falls through to the default case, and the default handler implements the desired behavior. It also allows watch faces to override certain behaviors: a watch face that really needs the MODE button can simply implement a case for `EVENT_MODE_BUTTON_UP` or `EVENT_MODE_LONG_PRESS`, and the default handler will not be called. 

At this time the PR only makes this change in the watch faces in the standard firmware. It should be backwards compatible, since all watch faces that don't implement the new handler currently *do* implement the default behaviors, just in their own cases in the switch statement. 

Anyway, opening this PR up for comment before making the change to the remaining watch faces.

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -218,6 +218,26 @@ void movement_illuminate_led(void) {
     }
 }
 
+bool movement_default_loop_handler(movement_event_t event, movement_settings_t *settings) {
+    (void)settings;
+
+    switch (event.event_type) {
+        case EVENT_MODE_BUTTON_UP:
+            movement_move_to_next_face();
+            break;
+        case EVENT_LIGHT_BUTTON_DOWN:
+            movement_illuminate_led();
+            break;
+        case EVENT_MODE_LONG_PRESS:
+            movement_move_to_face(0);
+            break;
+        default:
+            break;
+    }
+
+    return true;
+}
+
 void movement_move_to_face(uint8_t watch_face_index) {
     movement_state.watch_face_changed = true;
     movement_state.next_watch_face = watch_face_index;
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -229,7 +229,11 @@ bool movement_default_loop_handler(movement_event_t event, movement_settings_t *
             movement_illuminate_led();
             break;
         case EVENT_MODE_LONG_PRESS:
-            movement_move_to_face(0);
+            if (MOVEMENT_SECONDARY_FACE_INDEX && movement_state.current_watch_face == 0) {
+                movement_move_to_face(MOVEMENT_SECONDARY_FACE_INDEX);
+            } else {
+                movement_move_to_face(0);
+            }
             break;
         default:
             break;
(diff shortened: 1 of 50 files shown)


movement/movement.c
```diff
@@ -218,6 +218,26 @@ void movement_illuminate_led(void) {
...
+    (void)settings;
+
+    switch (event.event_type) {
+        case EVENT_MODE_BUTTON_UP:
+            movement_move_to_next_face();
+            break;
+        case EVENT_LIGHT_BUTTON_DOWN:
+            movement_illuminate_led();
+            break;
+        case EVENT_MODE_LONG_PRESS:
```
wryun, line 231: Probably could/should handle the MOVEMENT_SECONDARY_FACE_INDEX stuff here as well.

movement/template/template.c
```diff
@@ -76,7 +73,12 @@ bool <#watch_face_name#>_face_loop(movement_event_t event, movement_settings_t *
...
-            break;
+            // Movement's default loop handler will step in for any cases you don't handle above:
+            // * EVENT_LIGHT_BUTTON_DOWN lights the LED
+            // * EVENT_ALARM_BUTTON_UP moves to the next watch face in the list
+            // * EVENT_MODE_LONG_PRESS returns to the first watch face in the list
+            // You can override any of these behaviors by adding a case for these events to this switch statement.
+            return movement_default_loop_handler(event, settings);
     }
 
     // return true if the watch can enter standby mode. If you are PWM'ing an LED or buzzing the buzzer here,
```
wryun, line 84: This comment is not really accurate, AFAIK, since the movement code handles these cases automatically.

movement/movement.c
```diff
@@ -218,6 +218,26 @@ void movement_illuminate_led(void) {
...
 
+bool movement_default_loop_handler(movement_event_t event, movement_settings_t *settings) {
+    (void)settings;
+
+    switch (event.event_type) {
+        case EVENT_MODE_BUTTON_UP:
+            movement_move_to_next_face();
+            break;
+        case EVENT_LIGHT_BUTTON_DOWN:
+            movement_illuminate_led();
```
wryun, line 229: I think this could be confusing, because both this function and the previous function require the watch _not_ to sleep, but this is handled elsewhere. Maybe just a comment explaining why it's 'safe' to return true here, and amend the comment in `template.c` (see below).

wryun (approved):
:+1: nice. Huge fan of having all the default behaviour centralised, and this is an elegant solution to not changing the signature of the loop.

movement/movement.c
joeycastillo: Actually I think it was always a mistake for the watch to return false in these cases; `movement_move_to_next_face` sets a variable indicating that the watch face changed (which suppresses standby for one cycle) and `movement_illuminate_led`also sets a boolean that keeps the watch awake while the PWM is in operation. The `return false` in some watch faces was harmless (and confusing) but ultimately not necessary.

movement/movement.c
joeycastillo: Good call 👍

movement/template/template.c
joeycastillo: There is a typo here (`EVENT_ALARM_BUTTON_UP` should be `EVENT_MODE_BUTTON_UP `) but otherwise I think it is accurate? 

In context, this comment will appear in `my_watch_face.c`. If in the switch statement in that file, I define a case for `EVENT_MODE_LONG_PRESS`, the `default:` case will never be reached, and I will have overridden the behavior; the Movement code will not handle `EVENT_MODE_LONG_PRESS`.

movement/movement.c
wryun: Sorry, I should have been clearer - that was my point, that 'sleeping' is not necessary because it's handled. But then in the comment in template it says that if you mess with the LED you _do_ need to return false, which isn't true.

Err, I'm not making this better, am I?

Suggestion:

Comment here to say '// Safe to return true even though we don't want to sleep because movement forces us not to sleep'

And change template.c comment to mention that it's not necessary to return false in the LED case or the buzzer case as this is handled automatically.

movement/template/template.c
wryun: Sorry, my comment was, err, about the subsequent comment (not the one you added). I've tried to explain better in my other response.

movement/movement.c
joeycastillo: Aha, I get your point. I think the issue is that the documentation around messing with the LED is inartfully worded. The point I was trying to make is that if you're using the low-level watch library functions to interact with the LED or buzzer, you have to return false. Calling the higher-level Movement functions for interacting with the LED or buzzer does not require anything special.

I will resolve this by clarifying that note.

PR #184, opened 2023-01-20
alarm face: fix bell indicator not showing up on fridays

TheOnePerson:
This fixes a minor logic error which lead to a missing bell indicator on a friday when you have set an alarm for each workday.


joeycastillo:
\o/

PR #187, opened 2023-01-23
Activity face + Chirpy TX

gugray:
Contains the Activity face, which allows you to record acitivities like you would with a fitness watch.

The activities can be chirped out through the watch's piezo buzzer with the chirpy-tx transmitter that is added to movement/lib.

A separate chirpy_demo_face is included to illustrate the usage of chirpy-tx. It contains an option to chirp out Nanosec.ini if it exists on the watch, besides some hard-wired sample data.

Chirpy transmissions can be decoded with the [Sensor-Watch-Chirpy-RX](https://github.com/gugray/Sensor-Watch-ChirpyRX) web app. The app receives the binary stream, but it will also interpret recognized transmissions from the Activity face, or the values in Nanosec.ini.


movement/watch_faces/complication/activity_face.c
```diff
@@ -0,0 +1,714 @@
...
+    "  gYnn",
+    "  rOuu",
+    "SOCCEr",
+    " FOOTb",
+    " bALL ",
+    "  SKI ",
+};
+
+// Number of currently enabled activities (size of enabled_activities).
+static const uint8_t num_enabled_activities = 14;
```
neutralinsomniac, line 82: you can make this dynamic with something like:
`static const uint8_t num_enabled_activities = sizeof(activity_names) / sizeof(*activity_names);`

movement/watch_faces/complication/activity_face.c
```diff
@@ -0,0 +1,714 @@
...
+//   activities that you want to see in your watch.
+// You can also add new items to activity_names, but don't redefine or remove existing ones.
+
+// If a logged activity is shorter than this, then it won't be added to log when it ends.
+// This way scarce log slots are not taken up by aborted events that weren't real activities.
+static const uint16_t activity_min_length_sec = 10;  // 300
+
+// Supported activities. ID of activity is index in this buffer
+// W e should never change order or redefine items, only add new items when needed.
+static const char activity_names[][14] = {
```
neutralinsomniac, line 64: this should be:
`static const char activity_names[][6] = {`

movement/watch_faces/complication/activity_face.c
neutralinsomniac: oh actually I see that it's actually supposed to be *enabled* activities. in that case you can just do a `sizeof(enabled_activities)` (and move this to after the enabled_activities array too obviously)

movement/watch_faces/complication/activity_face.c
neutralinsomniac: actually, make that 6 a 7 to leave room for the terminating NULL. `watch_display_string()` will stop printing after reaching the end of the display, but sprintf() will happily copy bytes into buffers until reaching a NULL.

movement/watch_faces/complication/activity_face.c
gugray: Thanks, great point, I didn't know I could this in C. Fixed; pushing commit in a bit.

movement/watch_faces/complication/activity_face.c
gugray: Silly mistake, well spotted, thanks! Fixing in next commit.

movement/movement_config.h
```diff
@@ -29,12 +29,8 @@
 
 const watch_face_t watch_faces[] = {
     simple_clock_face,
-    world_clock_face,
```
neutralinsomniac, line 32: I would revert this file to its main branch state; if this gets merged as-is it'll overwrite the default faces.

gugray: good point! done.

joeycastillo:
I plan to merge this soon, but I want to spend a little more time playing with it! Alas the test setup I have right now isn't ideal for this. But I promise I'll get it merged soon — this is very, very cool!

PR #188, opened 2023-01-23
tarot_face: use direct character arrays instead of arrays of pointers to global strings

neutralinsomniac:
sorry for all these PR's; I just keep getting inspiration/ways to improve things from other people's faces and conversations in the discord.


neutralinsomniac:
This one saves 4 bytes per string

PR #189, opened 2023-01-25
add default loop handler to all current watch faces

neutralinsomniac:


joeycastillo:
Thanks so much for doing this! Wow you really got the grand tour of watch faces :)

neutralinsomniac:
heh I did! Got something useful out of it too! I had planned on writing a pomodoro face, but found the tomato face while doing this :)

PR #190, opened 2023-01-26
countdown_face: implement holding light to reset the currently edited timer value (and everything below it)

neutralinsomniac:
an alternative implementation could have us resetting *only* the currently edited value position and nothing else by replacing the `// intentional fallthrough` comments with `break;`'s


joeycastillo:
Looks good to me!

PR #195, opened 2023-01-28
ab_face

tslilc:
Activity/Break face. Essentially two countdown timers with auto-repeat and which are interleaved, as well the option for a number of laps.

I understand that there is already the very feature-ful `interval` face, so it would make sense not to include duplicate functionality. I created this because `interval` was too complex for my needs and didn't seem to have a "track number of times this has happened" mode.

### DIFF
--- a/movement/make/Makefile
+++ b/movement/make/Makefile
@@ -96,6 +96,7 @@ SRCS += \
   ../watch_faces/complication/morsecalc_face.c \
   ../watch_faces/complication/rpn_calculator_face.c \
   ../watch_faces/complication/ships_bell_face.c \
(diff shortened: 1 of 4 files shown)


buckket:
This is very similar to the [tomato face](https://github.com/joeycastillo/Sensor-Watch/blob/main/movement/watch_faces/complication/tomato_face.c), except for the laps. The tomato face only counts completed cycles.

tslilc:
thanks for taking the time to review this! Indeed, it's somewhere between tomato face and interval, the former doesn't have laps or dynamic setting of times, and the latter has all of this and more. I completely understand if we don't want to populate the repo with intermediate efforts. Thanks again :)

PR #196, opened 2023-01-28
Update counter_face.c: Non-blocking sound, silent option

Hein-NonesensE:
Hi, I have changed the sound to be non-blocking, so we can count faster than the sound is played. And I added an option to deactivate sound, for those of us who prefer silence.

I hope you like it and I hope it's up to standard as I'm not a pro(grammer). At least, it works on my watch :-)

### DIFF
--- a/movement/watch_faces/complication/counter_face.c
+++ b/movement/watch_faces/complication/counter_face.c
@@ -2,7 +2,6 @@
  * MIT License
  *
  * Copyright (c) 2022 Shogo Okamoto
- * Copyright (c) 2022 Jan H. Voigt
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
@@ -54,7 +53,7 @@ bool counter_face_loop(movement_event_t event, movement_settings_t *settings, vo
 
     switch (event.event_type) {
         case EVENT_ALARM_BUTTON_UP:
-            watch_buzzer_abort_sequence; //abort running buzzer sequence when counting fast
+            watch_buzzer_abort_sequence(); //abort running buzzer sequence when counting fast
             state->counter_idx++; // increment counter index
             if (state->counter_idx>99) { //0-99
                 state->counter_idx=0;//reset counter index
@@ -65,7 +64,7 @@ bool counter_face_loop(movement_event_t event, movement_settings_t *settings, vo
             }
             break;
         case EVENT_LIGHT_LONG_PRESS:
-            watch_buzzer_abort_sequence; 
+            watch_buzzer_abort_sequence(); 
             beep_on = !beep_on;
             if (beep_on){
             watch_set_indicator(WATCH_INDICATOR_SIGNAL);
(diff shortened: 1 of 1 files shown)


movement/watch_faces/complication/counter_face.c
```diff
@@ -48,12 +54,25 @@ bool counter_face_loop(movement_event_t event, movement_settings_t *settings, vo
 
     switch (event.event_type) {
         case EVENT_ALARM_BUTTON_UP:
+            watch_buzzer_abort_sequence; //abort running buzzer sequence when counting fast
```
neutralinsomniac, line 57: This call and the other call on line 68 need to be `watch_buzzer_abort_sequence();`

Without the parentheses you aren’t actually calling a function.

movement/watch_faces/complication/counter_face.c
```diff
@@ -75,22 +94,46 @@ bool counter_face_loop(movement_event_t event, movement_settings_t *settings, vo
...
+        i++;
+        sound_seq[i] = 6;
+        i++;
+    }
+    if (high_count > 1) {
+        sound_seq[i] = -2;
+        i++;
+        sound_seq[i] = high_count-1;
+    }
+    watch_buzzer_play_sequence((int8_t *)sound_seq, NULL);
```
neutralinsomniac, line 136: sound_seq is a stack variable, and this function goes out of scope as soon as `watch_buzzer_play_sequence()` returns. This means that while the audio is playing (asynchronously), sound_seq can be overwritten by any future function calls pushing things onto the stack. `sound_seq` should probably be static, but also keep in mind then that your initialization will only take effect on the first call of `beep_counter()` so you may want to `memset()` it.

Hein-NonesensE: Thanks for your feedback! I implemented the suggested changes in a new commit.

movement/watch_faces/complication/counter_face.c
```diff
@@ -38,6 +40,9 @@ void counter_face_setup(movement_settings_t *settings, uint8_t watch_face_index,
 
 void counter_face_activate(movement_settings_t *settings, void *context) {
     (void) settings;
+    if (beep_on){
```
Kistelini, line 43:
```suggestion
    if (beep_on) {
```

movement/watch_faces/complication/counter_face.c
```diff
@@ -48,12 +53,25 @@ bool counter_face_loop(movement_event_t event, movement_settings_t *settings, vo
...
     switch (event.event_type) {
         case EVENT_ALARM_BUTTON_UP:
+            watch_buzzer_abort_sequence(); //abort running buzzer sequence when counting fast
             state->counter_idx++; // increment counter index
             if (state->counter_idx>99) { //0-99
                 state->counter_idx=0;//reset counter index
             }
             print_counter(state);
+            if (beep_on) {
             beep_counter(state);
```
Kistelini, line 63:
```suggestion
                beep_counter(state);
```

movement/watch_faces/complication/counter_face.c
```diff
@@ -48,12 +53,25 @@ bool counter_face_loop(movement_event_t event, movement_settings_t *settings, vo
...
+            }
+            break;
+        case EVENT_LIGHT_LONG_PRESS:
+            watch_buzzer_abort_sequence(); 
+            beep_on = !beep_on;
+            if (beep_on){
+            watch_set_indicator(WATCH_INDICATOR_SIGNAL);
+            }
+            else {
+            watch_clear_indicator(WATCH_INDICATOR_SIGNAL);
```
Kistelini, line 73:
```suggestion
            if (beep_on) {
                watch_set_indicator(WATCH_INDICATOR_SIGNAL);
            } else {
                watch_clear_indicator(WATCH_INDICATOR_SIGNAL);
```

movement/watch_faces/complication/counter_face.c
```diff
@@ -75,22 +93,47 @@ bool counter_face_loop(movement_event_t event, movement_settings_t *settings, vo
 
 // beep counter index times
 void beep_counter(counter_state_t *state) {
-	
 	int low_count = state->counter_idx/5;
```
Kistelini, line 96: Indentation here was already mixed, maybe you can fix it, too?

PR #200, opened 2023-01-31
movement: add custom hourly chime tunes

neutralinsomniac:
Default movement behavior is completely unchanged unless one of the SIGNAL_TUNE_* defines are used. Battery impact is currently unknown, but the tunes are at least short (less than ~3 seconds long).

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -292,9 +292,13 @@ void movement_request_wake() {
 }
 
 void movement_play_signal(void) {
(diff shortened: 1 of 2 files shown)


neutralinsomniac:
Currently implemented tunes are the "found a secret" tune from Zelda, and the intro to the Mario theme. Additional tunes welcome.

neutralinsomniac:
Thinking about this more, I think it makes more sense to move it into its own watch face. Gonna refactor this and try again.

PR #202, opened 2023-02-01
Allow days in the future

kingannoy:
With this small change you can also count down to a important upcoming event!

This is my first time coding in C, so please double check everything! But in the emulator this seemed to work!


kingannoy:
I was a bit (OK, very) enthusiastic when I submitted this last night. My first time contributing a code change to a open source project!

But later I had a few reservations:
- I'm not sure this is the most elegant way of doing this. Maybe C has a operator for getting the difference between two numbers?
- The compiler showed a error about sprintf being vulnerable to overflowing? I think I already experienced that by setting a date in the future (the number gets very large then) but maybe I should see if I can also fix that?

For now this code runs on my watch and makes me happy 😃

joeycastillo:
I love it! We might need to change the year limiter to allow dates into the following year, but with this change if nothing else I now have my “Days until Christmas” counter locked down.

Thanks for the PR!!

PR #204, opened 2023-02-02
New face - Repetition Minute, fork of simple clock with an accessibillity feature

Jonas-fr:
A suggestion for a new face called "Repetition Minute". It's a direct fork of simple clock with an accessibility feature added : time reading via beeps (useful for blind people). Access it via a long push of the light button (on release).

Full description in the header file.

### DIFF
--- a/movement/make/Makefile
+++ b/movement/make/Makefile
@@ -96,6 +96,7 @@ SRCS += \
   ../watch_faces/complication/morsecalc_face.c \
   ../watch_faces/complication/rpn_calculator_face.c \
   ../watch_faces/complication/ships_bell_face.c \
+  ../watch_faces/clock/repetition_minute_face.c \
 # New watch faces go above this line.
 
 # Leave this line at the bottom of the file; it has all the targets for making your project.
--- a/movement/movement_faces.h
+++ b/movement/movement_faces.h
@@ -74,6 +74,7 @@
 #include "morsecalc_face.h"
 #include "rpn_calculator_face.h"
 #include "ships_bell_face.h"
+#include "repetition_minute_face.h"
 // New includes go above this line.
 
 #endif // MOVEMENT_FACES_H_
--- a/movement/watch_faces/clock/repetition_minute_face.c
+++ b/movement/watch_faces/clock/repetition_minute_face.c
@@ -1,0 +1,231 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2023 Jonas Termeau
+ *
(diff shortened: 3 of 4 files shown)


joeycastillo:
I love the concept! I would love to explore it a bit more though before merging. 

One question: is this based off an existing implementation of an accessible clock, or are we inventing it here? If it's the latter, I wonder if we could speed up the pace of the signal. I've worked a bit with engineering for screen reading tech like VoiceOver in my years as an iOS developer, and I get the sense that blind and low vision users of these technologies are comfortable with very fast readback speeds. Not saying we need to chirp out 10:00 in a tenth of a second, but I'm wondering if we could play with the implementation to make it quicker. 

I'd love to figure out a way to chirp out “10:47 AM” in a similar amount of time it would take a sighted person to glance at their watch. It might be as simple as speeding up the existing hour/quarter/minute beeps. Or perhaps you could break it up into digits? Ten high beeps, four midrange beeps, seven low beeps? Again I don't know, but if we're inventing an accessible clock here, I'd love to spend some extra time thinking about the UX and what would be most useful to the folks who would make the most use of it :)

Jonas-fr:
Thanks for your feedback! The minute repeater ("repetition minute" in French/Swiss) is one of the most expensive and rare complication and this one is based on Patek watches. See here for some videos of the complication in action : https://www.patek.com/en/company/savoir-faire/minute-repeaters.

Now regarding the speed I agree with you. While faithful with a very expensive (hundred of thousands to millions of euros/dollars) watch (which I find funny since the F-91 is one of the cheapest watch on the market), it's indeed a bit slow. The reason it's slow-ish in the first place is because of mechanical restrictions and because people who pay big money for a luxury watch love to take the time to enjoy the complications they're coming with. Counting too fast will ruin the effect and wore the mechanism of the watch.

That being said, from my point of view, there's some options for this face to address your point :
- make the whole counting faster, as simple as changing the timings in the play_xxx_chime functions
- introduce a configuration to this face that allows one to select the speed of the reading (normal (that one), fast, very fast), a bit more involved (to me) and it'd use (a little bit) more memory to save that config
- introduce a mode where it counts only the hours and the quarters and if the user needs the minutes also they have to long press again the alarm button (not too long after having listened to the hours/quarters counting) to get them
- introduce a new way to count, which I wouldn't be comfortable with, as it will stray further from the original counting mechanism this face pays homage to

Let me know what you think would be the best option and I'm also open to other contributors' point of view!

joeycastillo:
Reading your comments (and seeing the video of the Patek complications) I totally get it! And I think it makes sense to merge as is :)

PR #205, opened 2023-02-04
Sensor watch sailing face

Hein-NonesensE:
I have modified sailing_face, covering @niclashoyer 's suggestions from his initial PR and my wishlist.

Added features:
- stopwatch-like counter after start signal
- lap counter after start signal
- optional additional sounds at every minute, 30s, 10s countdown
- sound options (no sound, start only, all signals, all signals plus the additional sounds above)
- maximum starting time up to 10min (instead of 9)
- improved timing, display is no longer delayed by sound

I also have mitigated the issue I've talked about in Discord, where the countdown could be stopped by pressing a button while on another face. It's still theoretically possible, but I made the timeframe as short as I could and couldn't repoduce it since. Still, if anyone knows a real solution, I'd love to hear about it.

### DIFF
--- a/movement/watch_faces/complication/sailing_face.c
+++ b/movement/watch_faces/complication/sailing_face.c
@@ -1,6 +1,7 @@
 /*
  * MIT License
  *
+ * Copyright (c) 2023 Jan H. Voigt
  * Copyright (c) 2022 Wesley Ellis
  * Copyright (c) 2022 Niclas Hoyer
  *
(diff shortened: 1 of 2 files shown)


joeycastillo:
This looks good! And the fact that the changes are contained to one watch face makes it easy to merge :)

Having said that I don't know enough about sailing to really understand these changes... @niclashoyer what do you think, do these changes make sense to you? 

Thanks for the PR, and I'm stoked to get it merged! Such a cool use case for the watch.

niclashoyer:
@joeycastillo yes they make sense to me! Great to see this improved 🚀

PR #207, opened 2023-02-08
discgolf_face initial commit

RighteousLightning:
New watch face, keep score during rounds of discgolf or golf! 

Full description in C file.

### DIFF
--- a/movement/watch_faces/complication/discgolf_face.c
+++ b/movement/watch_faces/complication/discgolf_face.c
@@ -1,45 +1,36 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2023 Þorsteinn Jón Gautason
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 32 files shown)


movement/watch_faces/complication/discgolf_face.c
```diff
@@ -0,0 +1,319 @@
+#include <stdlib.h>
+#include <string.h>
+#include "discgolf_face.h"                                      // Remember to change number of courses in this file
+#include "watch.h"
+#include "watch_utility.h"
+
+/*
```
neutralinsomniac, line 7: Thank you for adding documentation! Would you mind moving this comment block into discgolf_face.h as per [template.h](https://github.com/joeycastillo/Sensor-Watch/blob/main/movement/template/template.h)? Thanks!

RighteousLightning: No problem! 
Comment block has been moved to the header file and I also added the license block to both files.

RighteousLightning:
Also, if anyone has an idea on how to make the high scores survive a battery change, that would be cool, currently it's initialized with all 99's at boot.

ostyn:
> Also, if anyone has an idea on how to make the high scores survive a battery change, that would be cool, currently it's initialized with all 99's at boot.

For export at least, this would be interesting:
https://github.com/gugray/Sensor-Watch-ChirpyRX

Issue #117, opened 2022-11-01
movement doesn't buffer serial input correctly


wryun:
movement.c uses `read` to grab something from the input, then processes it _immediately_, regardless of whether there's a newline in the buffer.

This means that if you use a normal terminal emulator (i.e. picocom), which sends characters immediately as they're typed - as opposed to Arduino's one, which sends a bunch of characters at once - every filesystem_process_command is called with a single character. And even using Arduino's, you can only send about 50 chars or so before the line is effectively broken at that point due to how fast the characters are transmitted (or some buffer size somewhere).

I believe the easiest fix is to implement an fgets-ish thing in the loop. Unfortunately, we can't use normal fgets at the moment (AFAIK), as it will block up the loop. Moreover, even if we were prepared to be a bit hacky by calling fgets only once there was input to read, AFAIK because select/poll isn't implemented there's no way to _detect_ if there's something is there without actually calling read(), and one is not supposed to mix read() with the normal stdio functions.

Note that from messing around with serial I/O I suspect there are also underlying issues with the serial implementation in TinyUSB, but this is the issue I'm 99% certain exists and is easily fixable.

wryun:
Just looked at the code in watch_private.c. There may be a few other issues:

- _read doesn't look like it can read NULL bytes, making binary transfer impossible (the output of tud_cdc_n_read isn't used)
- cdc_task always puts bytes at the start of the buffer, rather than appending to the buffer (what happens if no-one has called 'read' before the next TC0_Handler occurs?). Maybe the timings always work out ok...
- the buffer seems small to me and could easily overflow (but possibly relates to above... it would be nice at least to document what the timing constraints are here, I guess)

bademux:
test script
```python
import serial

ser = serial.Serial("/dev/ttyACM1",baudrate = 19200, timeout=1)

if ser.isOpen():

    ser.write("echo otpauth://totp/ACME%20Co:john.doe@email.com?secret=HXDMVJECJJWSRB3HWIZR4IFUGFTMXBOZ&issuer=ACME%20Co&algorithm=SHA1&digits=6&period=30 > totp_uris.txt".encode())
    print(ser.read(256).decode("ascii"))

    ser.write("ls".encode())
    print(ser.read(256).decode("ascii"))

    ser.write("cat totp_uris.txt".encode())
    print(ser.read(256).decode("ascii"))
ser.close()
```
and output is
```
$ python totp_test.py
[...]

PR #212, opened 2023-02-11
Some kind of serial buffering fix

wryun:
Addresses #117 

Removes the intermediate buffer entirely and make movement.c responsible for calling read. I'm not sure if anything bad happens if you _don't_ call tud_cdc_n_read, and this PR has problems with the initial display of '$' that I'm not sure how to work around well.

I'm raising it as a draft for comment since I lack confidence in its sanity.

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -67,6 +67,8 @@
 #define MOVEMENT_DEFAULT_GREEN_COLOR 0xF
 #endif
 
+#define SERIAL_BUFFER_SIZE 512
+
 #if __EMSCRIPTEN__
 #include <emscripten.h>
 #endif
@@ -424,6 +426,10 @@ static void _sleep_mode_app_loop(void) {
 }
 
 bool app_loop(void) {
+    static char serial_buffer[SERIAL_BUFFER_SIZE] = {0};
+    static int serial_buffer_index = 0;
+    static int prompt_shown = false;
+
     if (movement_state.watch_face_changed) {
         if (movement_state.settings.bit.button_should_sound) {
             // low note for nonzero case, high note for return to watch_face 0
@@ -520,7 +526,6 @@ bool app_loop(void) {
 
     // if we are plugged into USB, handle the file browser tasks
     if (watch_is_usb_enabled()) {
-        char line[256] = {0};
 #if __EMSCRIPTEN__
         // This is a terrible hack; ideally this should be handled deeper in the watch library.
         // Alas, emscripten treats read() as something that should pop up an input box, so I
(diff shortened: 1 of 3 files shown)


wryun:
IMO the alternative to doing it this 'stupid' way is to do something like what's in https://github.com/micropython/micropython/blob/d68e3b03b1053a6de0c7eb28f5989132c138364b/ports/samd/mphalport.c

watch-library/hardware/watch/watch_private.c
```diff
@@ -255,33 +255,22 @@ int _write(int file, char *ptr, int len) {
...
 
-static char buf[256] = {0};
-
 int _read(int file, char *ptr, int len) {
     (void)file;
-    int actual_length = strlen(buf);
-    if (actual_length) {
-        memcpy(ptr, buf, min(len, actual_length));
-        return actual_length;
+
```
wryun, line 260: This avoids the following issues:

- calling read more than once before cdc_task happened would give us the same data
- not calling read before cdc_task happened would silently drop data
- inability to read '\0'

movement/movement.c
```diff
@@ -533,15 +537,47 @@ bool app_loop(void) {
...
             tx = "";
         });
+        if (serial_buffer[0] != '\0') {
+            printf("%s", serial_buffer);
+            filesystem_process_command(serial_buffer);
+            serial_buffer[0] = '\0';
+        }
 #else
-        read(0, line, 256);
+        char *read_loc = &serial_buffer[serial_buffer_index];
```
wryun, line 553: This elaborate buffering dance handles arbitrary amounts of input at a time. I've been using:

```picocom /dev/ttyACM0 --imap lfcrlf```

File: watch-library/hardware/watch/watch_private.c [hunk 3]
```diff
@@ -245,43 +245,35 @@ void _watch_enable_usb(void) {
...
 // this function ends up getting called by printf to log stuff to the USB console.
 int _write(int file, char *ptr, int len) {
+    int written = 0;
     (void)file;
-    if (hri_usbdevice_get_CTRLA_ENABLE_bit(USB)) {
-        tud_cdc_n_write(0, (void const*)ptr, len);
-        tud_cdc_n_write_flush(0);
-        return len;
+
+    while (written < len) {
```
wryun, line 251: Previously, if you tried to 'write' something large it would write whatever it could (tud_cdc_n_write) and then claim it had written the whole buffer.

I think it should have been enough to just properly return what was written here (i.e. the return value of tud_cdc_n_write), but in practice this failed on catting large files (just cutting off the file midway, as opposed to the earlier behaviour of cutting off each line). Suspect there's some kind of buffering in newlib that isn't behaving the way I'd expect, but doing the stupid loop here at least seems to work.

watch-library/hardware/watch/watch_private.c
```diff
@@ -245,43 +245,35 @@ void _watch_enable_usb(void) {
...
 // this function ends up getting called by printf to log stuff to the USB console.
 int _write(int file, char *ptr, int len) {
+    int written = 0;
     (void)file;
-    if (hri_usbdevice_get_CTRLA_ENABLE_bit(USB)) {
-        tud_cdc_n_write(0, (void const*)ptr, len);
-        tud_cdc_n_write_flush(0);
-        return len;
+
+    while (hri_usbdevice_get_CTRLA_ENABLE_bit(USB) && written == 0) {
```
wryun, line 251: Previously, if you tried to write something large it would write whatever it could (tud_cdc_n_write) and then claim it had written the whole buffer.

Unfortunately, https://github.com/bminor/newlib/blob/5192d5ea51dee786ba1cb97b216bd40115dbbf20/newlib/libc/stdio/fflush.c#L225 demands that we actually write something othewise it's an error, hence there's a busy loop here waiting for our buffer to clear. This is presumably horribly inefficient, but I'm not sure how to do it the right way.

wryun: So, experimenting with do both USB CDC and mass storage at the same time, this blocking operation causes us to freeze up if we spam debug statements in the USB mass storage callbacks (probably because we're already busy with mass storage and can't write CDC things at the same time).

Not sure whether the lesson is 'don't do silly things in callbacks' or 'we need an intermediate buffer here' or ...

watch-library/hardware/watch/watch_private.c
```diff
@@ -245,43 +245,40 @@ void _watch_enable_usb(void) {
...
-        tud_cdc_n_write(0, (void const*)ptr, len);
-        tud_cdc_n_write_flush(0);
-        return len;
+
+    // It's important that this call blocks, as returning 0 bytes
+    // written causes newlib to fail the flush and throw
+    // away any data in the buffer.
+    // Note that it's fine not to write all the bytes, it's just
+    // that we have to write _something_.
+    while (tud_cdc_connected() && written == 0) {
```
wryun, line 256: I'm now wondering if it's a reasonable approach to just try to do this for some amount of time _then_ fail to avoid deadlocks. Of course, I could put an intermediate buffer in, but tiny USB already has a buffer here (which we could increase...).

wryun:
@WesleyAC - I realise this is not right, but I think it's better than the current situation (where the serial is IMO close to unusable). I'd moved it out of draft in hope :)

wryun:
Closed since #344 fixes this properly.

PR #213, opened 2023-02-12
Fix cat command

wryun:
If we ran into a heap that wasn't filled with '\0', fun things.


joeycastillo:
very fun! Good catch, thanks for this :)

PR #220, opened 2023-02-24
Adding Breathing complication

berndpl:
Breathing is a complication for guiding boxed breathing sessions.

Boxed breathing is a technique to help you stay calm and improve concentration in stressful situations. Here is a [demo](https://youtube.com/shorts/-MViA2ByNQc)

I appreciate any feedback for improving the code. Also if you feel this is not a fit for the project feel free to discard.


movement/watch_faces/complication/breathing_face.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include "movement.h"
+
+typedef struct {
+    uint8_t current_stage;
+    bool animating;
+    bool sound_on;
+} breathing_state_t;
+
+void beep_in (void);
```
wryun, line 36: Because these functions are not used outside breathing_face.c, it's unnecessary to have them in the '.h' file. It's also probably best to define them as 'static' in the '.c' file so the symbols can't clash with anything.

movement/watch_faces/complication/breathing_face.c
```diff
@@ -0,0 +1,198 @@
...
+
+            if (state->sound_on == true) {
+                watch_set_indicator(WATCH_INDICATOR_BELL); 
+            } else {
+                watch_clear_indicator(WATCH_INDICATOR_BELL); 
+            }
+
+            switch (state->current_stage)
+            {
+            case 0: { watch_display_string("Breath", 4); if (state->sound_on) beep_in(); } break;
```
wryun, line 134: It's fine like this, but I would be tempted to put this data in an array. In fact, I think you could represent the whole set of steps as something like:

```
struct {
    char *str;
    void (*sound_fn)(void);
} stages[] = {
    {"Breath", beep_in},
    {"In    3", noop},
    ...
};
```

But I'm probably just saying this because I don't like switch statements that much :)

movement/watch_faces/complication/breathing_face.c
```diff
@@ -0,0 +1,198 @@
...
+        };
+        for(size_t i = 0, count = sizeof(notes) / sizeof(notes[0]); i < count; i++) {
+            watch_buzzer_play_note(notes[i], durations[i]);
+        }
+}
+
+void beep_out_hold (void) {
+        const BuzzerNote notes[] = {
+            BUZZER_NOTE_C4,
+            BUZZER_NOTE_REST * 2,
```
wryun, line 105: I think this `*2` should be applied to `NOTE_LENGTH` instead.

wryun (commented):
Looks good! I'm just a random commenter, so I can't actually approve/merge this (and don't take anything I write too seriously).

movement/watch_faces/complication/breathing_face.c
berndpl: I like this. I actually wanted remove the data from the beeping portion, too. But that's where my non-existing proficiency in C stopped me. No completely sure where I need the pointers and how to deal with arrays in that context. Still need to learn that. Thanks.

movement/watch_faces/complication/breathing_face.c
berndpl: Yes. Good catch. Thanks.

joeycastillo:
Thanks so much for making this, and sorry for the delay in getting it merged in!

PR #224, opened 2023-03-06
Timer Face: Advanced countdown face with presets

TheOnePerson:
This is a modified version of the countdown watch face, adding two features:
- choosing from up to nine preset timer lengths
- auto-repeat a timer

I have thought about making a PR for the existing countdown face, but went with proposing this as a separate watch face. I had to alter the user interface (button functions) significantly to implement esp. the first feature, so it might be the best to let the user choose if he/she goes with the basic countdown face or this more advanced timer face.

How to use it:

The timer watch face provides the functionality of starting a countdown by choosing one out of nine programmable timer presets. A timer/countdown can be 23 hours, 59 minutes, and 59 seconds max. A timer can also be set to auto-repeat, which is indicated by the lap indicator.

How to use in NORMAL mode:
  - Short-pressing the alarm button cycles through all pre-set timer lengths.
  - Long-pressing the alarm button starts the timer.
  - Long-pressing the light button initiates settings mode.

How to use in SETTINGS mode:
  - There are up to nine slots for storing a timer setting. The current slot is indicated by the number in the upper right-hand corner.
  - Short-pressing the light button cycles through the settings values of each timer slot in the following order: hours - minutes - seconds - timer repeat
  - Short-pressing the alarm button alters the current settings value.
  - Long-pressing the light button resumes to normal mode.

### DIFF
--- a/movement/make/Makefile
+++ b/movement/make/Makefile
@@ -97,6 +97,7 @@ SRCS += \
   ../watch_faces/complication/rpn_calculator_face.c \
   ../watch_faces/complication/ships_bell_face.c \
   ../watch_faces/complication/habit_face.c \
+  ../watch_faces/complication/timer_face.c \
 # New watch faces go above this line.
 
 # Leave this line at the bottom of the file; it has all the targets for making your project.
--- a/movement/movement_faces.h
+++ b/movement/movement_faces.h
@@ -75,6 +75,7 @@
 #include "rpn_calculator_face.h"
 #include "ships_bell_face.h"
 #include "habit_face.h"
+#include "timer_face.h"
 // New includes go above this line.
 
 #endif // MOVEMENT_FACES_H_
--- a/movement/watch_faces/complication/timer_face.c
+++ b/movement/watch_faces/complication/timer_face.c
@@ -1,0 +1,362 @@
+/*
+ * MIT License
+ *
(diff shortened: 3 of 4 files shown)


Niehztog:
Hi @TheOnePerson , great PR, I have already flashed it onto my watch and am actively using it. I have been waiting for exactly this feature as I often use countdowns for different things. What do you think, wouldn't it make sense to show the active slot number in normal mode as well? This would help me when browsing the slots.

TheOnePerson:
Thanks, @Niehztog fir the feedback. I have actually been thinking about this, but had the feeling that showing the slot number in normal mode does not provide useful information. Since you can see the countdown length, I felt like showing the slot number could more be a source for confusion. But I am not decided on this at all. In what way would the number help you?

Niehztog:
It would provide a means for orientation of where in the slot queue I currently am. Of course it is correct that seeing the slot duration is "sufficient", but for example if I know the slot I use for rice cooking is number 7, I can see how many button presses are needed in order to get to 7. Or maybe I forgot the exact amount of time for boiling eggs (or my daily meditation practice or whatever), or I have two very similar slot durations, but still remember the slot number.

TheOnePerson:
Those are good points! I will implement this change in the next days. 👍

TheOnePerson:
Done, the face now shows the current timer slot number also in normal mode (and not only in settings mode).

PR #230, opened 2023-03-14
Aperture priority light meter face

enthdegree:
Added a lightmeter. See `movement/watch_faces/sensor/lightmeter_face.c` for usage details


enthdegree:
Oops, I can already tell I did something wrong: the license on the KiCad OPT3001 model is a bad one and I need to get rid of it. 

Can anyone offer advice here?

PR #231, opened 2023-03-25
Advanced Location Management

randogoth:
Two watch faces Place and Places.

The first one is an editor for the watch location that supports several different coordinate notations: 
- Latitude & Longitude in both decimal format and Degrees, Minutes, Seconds. 
- Open Location Code (aka. PlusCode) and Geohash fixed to 10 digits.

The second one is an optional addition to the first and allows to save up to 5 different locations with a custom name and quickly swap between them.

-------> More details in the header files.

This also attempts to establish a file <place.loc> as a standard for having the location optionally stored in a higher precision than the two decimal digits of the location saved in the watche's register.


randogoth:
fixed all warnings

randogoth:
bumping this due to #378

Issue #246, opened 2023-05-17
Timer face loop broken.


ysftaha:
# Summary
The timer face breaks after the second loop.
# Steps to reproduce
1. set a timer for 15 minutes
2. set to loop
# Result
After the second loop, the face turns into nonsense
![IMG_3459](https://github.com/joeycastillo/Sensor-Watch/assets/42289530/b821a9e3-8312-41d1-8071-3b84b296b9c8)
Image by @TheOnePerson

TheOnePerson:
I created PR #247 and tested it on my wrist. Looks good then.

PR #247, opened 2023-05-18
Timer face: fix error with repeating timers

TheOnePerson:
This PR fixes issue #246 and makes an additional minor change: When you start a timer, a shot beep will sound, even if you did not activate "beeps" in the watch settings.


joeycastillo:
Sweet!

PR #250, opened 2023-05-29
Face for tracking the menstrual cycle

jokomo24:
I discovered the Casio F-91W through my partner, appreciated the retro analog aesthetic of the watch, and got one for myself. Soon afterward I discovered the Sensor Watch project and ordered two boards, thank you!! I introduced the Sensor Watch to my partner who inquired whether she could track her menstrual cycle. So I decided to implement a menstrual cycle watch face that also calculates the peak fertility window using The Calendar Method.


joeycastillo:
This is such a great idea, thanks for making this! And monumental apologies for the delay in reviewing it. 

All looks good, except for one warning that I need your guidance to resolve. The switch statement on line 326, handing the alarm long press, can fall through. I'm not sure if that is intentional or not. Can you resolve by either adding a `break;` at the end of that case, or if the fall-through is intentional, by adding a `// fall through` comment where the break would be? Either of these things will resolve the warning. 

Thanks again!

jokomo24:
Of course and thank you for your hard work, this project is awesome! My apologies as well for taking so long to respond, I just completed a move and a job search so times were a bit hectic.

But yes, nice catch! I did not intend to let this EVENT_ALARM_LONG_PRESS case to fall through. Also, I uncommented my use of the beep function for "production". The simulator would freeze during testing of the watch face if I did not comment it out and I could not determine exactly why.

Thanks again for your feedback and the chance to contribute!

jokomo24:
Now that I've gotten this face on a physical watch, I've noticed a few bugs. Mainly, the value for "Period in # days" and for "Average Cycle" becomes zero after a period of time. I will be looking into it!

jokomo24:
I've identified that the issue occurs when reseting the watch face using the reset page. So the bug must reside in my reset function...

movement/movement_config.h
CarpeNoctem, line 1: Don't think we generally want to update this file for most pull requests, unless we want to change the default set of faces. (Awesome work otherwise, though! Great face idea.)

movement/movement_faces.h
CarpeNoctem, line 1: I'm curious - did you use the python script to generate your face boilerplate from template? Usually it puts new faces at the bottom of the list in this and the Makefile (not that the order here matters).

CarpeNoctem:
> I've identified that the issue occurs when reseting the watch face using the reset page. So the bug must reside in my reset function...

Nice work on this face! Just checking - has this all been resolved as of c1ac01d3d57d6e4670bf8d9e1e177a92a5cb1e92?

movement/movement_config.h
jokomo24: Ah yes, of course! I changed the order for easy testing and it got included because I used a git add all cmd, my bad! Thank you!

CarpeNoctem: I did the same. Very easy to add all out of habit. :)

movement/movement_faces.h
jokomo24: I did not, I don't think I was aware of it at the time but I will next time! I believe I used one of the other faces as a template but can't remember which at the moment.

jokomo24:
> > I've identified that the issue occurs when reseting the watch face using the reset page. So the bug must reside in my reset function...
> 
> Nice work on this face! Just checking - has this all been resolved as of [c1ac01d](https://github.com/joeycastillo/Sensor-Watch/commit/c1ac01d3d57d6e4670bf8d9e1e177a92a5cb1e92)?

Yes, this has been resolved! Was a simple typo mistake where I used `watch_store_backup_data(state->dates.reg, state->backup_register_cy)` instead of `watch_store_backup_data(state->cycles.reg, state->backup_register_cy)`.

Also, in `reset_tracking()`, I tried to clear the entire register using `state->dates.reg = 0` and `state->cycles.reg = 0` but that did not work as expected, so I went back to setting the bits to 0 explicitly just like in `menstrual_cycle_face_setup()`.

Thank you for the review and acknowledgements!

movement/movement_config.h
```diff
@@ -28,6 +28,7 @@
 #include "movement_faces.h"
 
 const watch_face_t watch_faces[] = {
+    menstrual_cycle_face,
```
WesleyAC, line 31: This line should get removed — we don't want to change the default set of faces.

movement/watch_faces/complication/menstrual_cycle_face.c
```diff
@@ -0,0 +1,471 @@
...
+    uint8_t cycle_length = total_days_tracked(state) - (unix_prev_period / SECONDS_PER_DAY);
+
+    // Update the shortest or longest cycle length if necessary
+    if (cycle_length < state->cycles.bit.shortest_cycle)
+        state->cycles.bit.shortest_cycle = cycle_length;
+    else if (cycle_length > state->cycles.bit.longest_cycle)
+        state->cycles.bit.longest_cycle = cycle_length;
+}
+
+void menstrual_cycle_face_setup(movement_settings_t *settings, uint8_t watch_face_index, void ** context_ptr) {
```
WesleyAC, line 253: This triggers a warning, since the `settings` parameter isn't used. To silence the warning, you can add the line:

```c
(void) settings;
```

next to the `(void) watch_face_index;` that's already there.

movement/watch_faces/complication/menstrual_cycle_face.c
```diff
@@ -0,0 +1,471 @@
...
+            break;
+        case average_cycle:
+            sprintf(buf, "%2d", state->cycles.bit.average_cycle);
+            watch_display_string(buf, 2);
+            break;
+        case peak_fertility_window:
+            if (event.subsecond % 5 && state->dates.reg) { // blink active for 3 quarter-seconds
+                first_day_fert = get_day_pk_fert(state, first_day);
+                last_day_fert = get_day_pk_fert(state, last_day);
+                sprintf(buf, "Fr%2d To %2d", first_day_fert, last_day_fert); // From: first day | To: last day
```
WesleyAC, line 431: This line gives the following warning:

```
../watch_faces/complication/menstrual_cycle_face.c: In function 'menstrual_cycle_face_loop':
../watch_faces/complication/menstrual_cycle_face.c:431:40: warning: '%2d' directive writing between 2 and 3 bytes into a region of size between 2 and 3 [-Wformat-overflow=]
../watch_faces/complication/menstrual_cycle_face.c:431:30: note: directive argument in the range [0, 255]
../watch_faces/complication/menstrual_cycle_face.c:431:17: note: 'sprintf' output between 11 and 13 bytes into a destination of size 11
```

This isn't actually a problem, since the day number will always be between 1 and 31, but it's good to get rid of the warning anyways — a simple way to do that is to just write:

```c
sprintf(buf, "Fr%2d To %2d", first_day_fert % 100, last_day_fert % 100);
```

movement/watch_faces/complication/menstrual_cycle_face.h
```diff
@@ -0,0 +1,80 @@
...
+#ifndef MENSTRUAL_CYCLE_FACE_H_
+#define MENSTRUAL_CYCLE_FACE_H_
+
+#include "movement.h"
+
+typedef struct {
+    // Store the date of the 'first' and the total cycles since to calulate and store the average menstrual cycle.
+    // Store the date of the previous, most recent, period to calculate the cycle length.
+    // Store the shortest and longest cycle to calculate the fertility window for The Calender Method.
+    // NOTE: Not thrilled about using two registers, but could not find a way to perform The Calender Method
```
WesleyAC, line 34: It might be preferable to store this data in LFS instead of in the backup registers, just since there are a limited number of backup registers and a lot of LFS space. Then again, currently we have unused backup registers, so it seems fine to merge this as-is and change it to use LFS later if we want.

#322 has a example of how to use LFS

WesleyAC (changes requested):

joeycastillo:
Sincere apologies for the delay in merging your pull request, and once again thank you so much for making this!

PR #252, opened 2023-06-25
Fix simulator and hardware divergence in callback handling

nategraf:
When using the simulator, I encountered cases where the light would become stuck on, and the watch
would be unresponsive. In particular, this would occur when pressing the light button on the
sunrise sunset watch face.

I appears that this is caused by a divergence in out the callback mask is intepretted by the
hardware interface, and in the simulator in the following function.

```c
void watch_rtc_disable_matching_periodic_callbacks(uint8_t mask)
```

In particular, a mask of `0xFE` is intended to disable all except the 128hz cvallback at index 0,
but instead disables all except the 1hz callback at index 7 in the simulator.

Based on by reading through and debugging the code in the simulator, this should fix a divergence
between then hardware and simulator, but I am not 100% I am correct here. I see that @a2 originally
included this code in https://github.com/joeycastillo/Sensor-Watch/pull/43, so I'd love your eyes
here.


nategraf:
@a2 and/or @joeycastillo do you have a moment to look at this PR?

.gitignore
```diff
@@ -2,6 +2,7 @@
 **/build-sim/
 *.b#*
 *.bin
+.cache/
```
WesleyAC, line 5: This seems like it was added by accident, right?

WesleyAC (changes requested):
Other than the presumably accidental `.gitignore` change, this looks good. A frequency of 128 means that PER0 gets set, which is `tick_callbacks[0]`. `0xFE` should leave PER0 be but clear `PER1`-`PER7` as per the datasheet. `0xFE & (1 << 1) != 0` correctly does that, being false in the case of i=0 and true in the case of 1-7. Looks like the existing code just got this backwards, a understandable mistake.

.gitignore
nategraf: Reverted that change :+1:

PR #253, opened 2023-07-04
Counter face: change sound to non-blocking, add option to deactivate sound

Kistelini:
This is the same as https://github.com/joeycastillo/Sensor-Watch/pull/196 including my suggestions.


joeycastillo:
Very nice! Just tried it out and it kept up with all my button mashing. 

Thanks for this!

joeycastillo:
Closes #196

PR #255, opened 2023-07-07
Fix lengthBytesUTF8 not a function

austenadler:
I was not able to run the simulator due to `lengthBytesUTF8 not a function` error in the console being spammed. Related: https://github.com/emscripten-core/emscripten/issues/16098

I do not know what caused this error, but someone in discord had the same problem so might be more widespread than just myself.


joeycastillo:
Great catch, thanks for this!

PR #256, opened 2023-07-10
Wyoscan .5 hz watchface

wjhrdy:
This is a recreation of the Wyoscan watch, which was a $175 watch in 2014. It was an f-91w pcb replacement.

https://github.com/joeycastillo/Sensor-Watch/assets/1795778/e07f0ed1-e328-4337-a654-fa1ee65d883f

Here is more information on it:
https://artmetropole.com/shop/11460

a demo of what it looks like:
https://www.o-r-g.com/apps/wyoscan

8 frames per number * 6 numbers + the trailing 16 frames = 64 frames
at 32 frames per second, this is a 2-second cycle time or 0.5 Hz.

It is giving me a stack overflow after about 2.5 cycles of the time display in the emulator but it works fine on the watch.

image

I'd like to make something for the low energy mode, but I haven't thought about how that might work, right now it just freezes in low energy mode until you press the 12-24HR button.

### DIFF
--- a/movement/watch_faces/clock/wyoscan_face.c
+++ b/movement/watch_faces/clock/wyoscan_face.c
@@ -57,7 +57,8 @@ static char *segment_map[] = {
     "AFBGECD",  // 8
     "AFBGCD"    // 9
 };
-static int32_t clock_mapping[6][7][2] = {
+
+const static int32_t clock_mapping[6][7][2] = {
     // hour 1
     {{1,18}, {2,19}, {0,19}, {1,18}, {0,18}, {2,18}, {1,19}},
     // hour 2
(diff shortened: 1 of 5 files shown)


movement/watch_faces/clock/wyoscan_face.c
```diff
@@ -0,0 +1,198 @@
...
+
+static bool watch_colon_animation_is_running(void) {
+    return colon_interval_id != -1;
+}
+
+static void watch_start_colon_animation(uint32_t duration) {
+    if (colon_interval_id != -1) return;
+
+    colon_state = true;
+    colon_interval_id = emscripten_set_interval(watch_invoke_colon_callback, (double)duration, NULL);
```
wryun, line 96: I think it might just be this line that's causing you problems, though I'm not entirely sure how. Why are you calling emscripten here?

wjhrdy: I'm just tried taking this out and it still stack overflows.

I'm calling it because it seems to be the way you do a blinking animation in the emulator.

wjhrdy:
CleanShot 2023-07-10 at 23 52 27

it seems like something in the js is calling the setinterval function and that is making it fail looking at the stack trace.

wjhrdy:
I funally put it on my watch and it is not crashing. yay

wjhrdy:
This is no longer a draft.

joeycastillo:
This is SO MONUMENTALLY COOL! My one desire would be for it to read the settings and display a 12-hour clock when the wearer has that preference set, but I'm not going to let that block merging this insanely cool watch face. 

As for low energy mode, I guess it could fall back to just displaying the time as normal, but again, I think that can be handled in a separate PR. 

Thanks so much for making this!!

PR #257, opened 2023-07-16
Face settings: A fairly easy way to store watch face data persistently

TheOnePerson:
In this PR, I have put together a solution for persting data of a watch face trying to keep the logic for storing and reading the data files as encapsulated as possible: `file_settings.c` & `file_settings.h` implementing two functions `face_settings_init()` and `face_settings_save()`.

- You can use the `face_settings_init()` function to initialize the details about the data structure belonging to your particular watch face. The function takes an identifying string for your watch face (must be unique) and a pointer to the data structure that will be saved to flash storage. You can also specify two callback functions (see below). It also checks whether a data file for your watch face already exists and if so, it loads it into the data structure.
- There is also basic logic for dealing with different versions of a watch face and possible changes in data structures. The data scheme version number is used to check if there are data files for previous versions present. If so, the old structure is being loaded to a temporary buffer and the version callback function is invoked, so you have a chance to transfer information from the previous version to the current one. The file belonging to the older data scheme version is erased afterwards (unless you specify otherwise by setting the `erase_file` variable to `false`).
- The format of the data files is pretty straightforward: The naming scheme is `data_[hash of identifier string]_[data scheme version]`. The file's content is a binary copy of the data structure.
- Whenever the watch face wants to persist it’s settings, it can call `face_settings_save()`. The current settings will only be saved when it’s contents have changed compared to the last call of `face_settings_save()`. This comparison is done by calculating a (hopefully) unique 4-byte hash value over the content of the data structure.
- Often, it is not feasible to simply write out the whole data structure as it is. For example, if the structure contain temporary properties that do not need to be saved. To handle this, and to further minimize the writing operations you can specify a 'save callback function' that will be invoked right before this hash comparison is done. Within this callback function, you can set irrelevant properties to a default value, so that changes in these properties do not lead to writing out a new data file.

The implementation of persistent storage for the **alarm face** looks like this:
- add `#include "face_settings.h"`, of course
- in `alarm_face_setup()` there is only one line added:
`face_data_init("alarm_face", 0, *context_ptr, sizeof(alarm_state_t), _face_save_data, NULL);`
- `_face_save_data()` is the function dealing with 'cleaning up' the data structure before saving it. It sets some properties to their defaults and looks like this:
```/// @brief default some data before saving to flash storage
static void _face_save_data(void *context) {
    alarm_state_t *state = (alarm_state_t *)context;
    state->alarm_idx = 0;
    state->alarm_playing_idx = 0;
    state->alarm_quick_ticks = false;
    state->alarm_handled_minute = -1;
    // default all one time alarms
    for (uint8_t i = 0; i < ALARM_ALARMS; i++) {
        if (state->alarm[i].day == ALARM_DAY_ONE_TIME) {
            state->alarm[i].day = ALARM_DAY_EACH_DAY;
            state->alarm[i].beeps = 5;
            state->alarm[i].pitch = 1;
            state->alarm[i].enabled = false;
            state->alarm[i].minute = 0;
            state->alarm[i].hour = 0;
        }
    }
}
```
- finally, the data is being checked for storing every two hours. This is done in `alarm_face_wants_background_task()` like so: `if (now.unit.minute == 0 && (now.unit.hour % 2) == 0) face_data_save(state);`

So far, everything looks fine in the emulator. I did not test it on hardware yet, but I guess I will give it a try later today.

P.S.: I had to heavily alter the filesystem.c, too, so that it now uses the browser's localStorage to save and load files. This is due to easier debugging and to get persistence at all in the emulator.

### DIFF
--- a/movement/face_settings.c
+++ b/movement/face_settings.c
@@ -86,20 +86,12 @@ bool face_data_init(const char* watch_face_identifier,
     face_data_details_t *face_data_details = _get_face_data_details(context);
     if (face_data_details == NULL) return false;
 
-    printf("face_data_init: callbacks %d, %d\n", save_callback, version_callback);
-
     // set face data details
     face_data_details->schema_version = schema_version;
     face_data_details->identifier_hash = DJBHash(watch_face_identifier, (uint16_t)strlen(watch_face_identifier));
     face_data_details->context_hash = 0;
     face_data_details->context_length = context_length;
     face_data_details->save_callback = save_callback;
-    printf("face_data_init: details pointer %d, identifier hash %08x, content hash %04x, length %d, context pointer %d, version callback %d\n",
-        face_data_details,
-        face_data_details->identifier_hash,
-        face_data_details->context_hash, 
-        face_data_details->context_length,
-        face_data_details->context_data, version_callback);
 
     // look for data file (current schema version)
     char filename[FACE_DATA_FILENAME_LEN];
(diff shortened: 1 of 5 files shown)


TheOnePerson:
Set this to draft until I had the chance to test it on the watch.

TheOnePerson:
Had it on my wrist and tested it several timers. Looks good to me.

TheOnePerson:
Converted this to draft again: I feel like I should simplify the data structure for holding face’s settings details a little bit, thus saving 4 bytes per entry. Will take care of it over the next couple of days.

TheOnePerson:
This is ready for now. If this works well (and does not wear out the flash storage) watch faces could use this to store their settings permanently.

TheOnePerson:
I had to make a small change to the preferences face. When the alarm face is present, too, I may happen that the preferences face restores movement's settings after the alarm face has restored its face settings, resulting in blanking out the alarm indicator erroneously.
Tbh, the whole construct of persisting movement settings from the preferences face feels a bit dodgy. The cleaner approach would be to move the logic for saving and restoring the data right into movement.c. But I hesitated to change this very central piece of code without feedback from @joeycastillo.

PR #258, opened 2023-07-20
fix nanosec_face correction_profile bounds from going negative

analogist:
The new nanosec_face has a `nanosec_state.correction_profile` of type `int8_t`, and can turn negative when ALARM is pressed from profile `P0`. From there the setting screen will display a cryptic `P-`, until the modulo arithmetic brings the profile back around to P0 again.

This makes two minor bugfixes to enforce bounds on `correction_profile`:

- roll around the `correction_profile` back to the highest profile when ALARM is pressed from P0 (based on `#define nanosec_profile_count`)
- prevent long press of ALARM from decrementing by 50 while on the profile screen - instead the behavior should still be to decrement by 1.


joeycastillo:
This makes sense, thank you for the PR!

PR #259, opened 2023-07-23
Game: Higher or Lower

Chris-E-J-Ellis:
A take at an implementation of the "Higher or Lower" or "Play Your Cards Right" game. I've tested it on the emulator, and currently have it running on my Sensor Watch; I'm not entirely convinced it's fun, but it seems to be working as intended =D

Dumping description from the header file below:

## Higher-Lower game face
======================

A game face based on the "higher-lower" card game where the objective is to correctly guess if the next card will
be higher or lower than the last revealed cards.

### Game Flow
- When the face is selected, the "Hi-Lo" "Title" screen will be displayed, and the status indicator will display "GA" for game
- Pressing `ALARM` or `LIGHT` will start the game and proceed to the "Guessing" screen
  - The first card will be revealed and the player must now make a guess
  - A player can guess `Higher` by pressing the `LIGHT` button, and `Lower` by pressing the `ALARM` button
  - The status indicator will show the result of the guess: HI (Higher), LO (Lower), or == (Equal)
  - There are five guesses to make on each game screen, once the end of the screen is reached, a new screen
    will be started, with the last revealed card carried over
  - The number of completed screens is displayed in the top right (see Scoring)
- If the player has guessed correctly, the score is updated and play continues (see Scoring)
- If the player has guessed incorrectly, the status will change to GO (Game Over)
  - The current card will be revealed
  - Pressing `ALARM` or `LIGHT` will transition to the "Score" screen
- If the game is won, the status indicator will display "WI" and the "Win" screen will be displayed
  - Pressing `ALARM` or `LIGHT` will transition to the "Score" screen
- The status indicator will change to "SC" when the final score is displayed
  - The number of completed game screens will be displayed on using the first two digits
  - The number of correct guesses will be displayed using the final three digits
  - E.g. "13: 063" represents 13 completed screens, with 63 correct guesses
- Pressing `ALARM` or `LIGHT` while on the "Score" screen will transition to back to the "Title" screen

### Scoring
- If the player guesses correctly (HI/LO) a point is gained
- If the player guesses incorrectly the game ends
  - Unless the revealed card is equal (==) to the last card, in which case play continues, but no point is gained
- If the player completes 40 screens full of cards, the game ends and a win screen is displayed

### Misc
The face tries to remain true to the spirit of using "cards"; to cope with the display limitations I've arrived at
the following mapping of card values to screen display, but am open to better suggestions:

```
| Cards   |                          |
|---------|--------------------------|
| Value   |2|3|4|5|6|7|8|9|10|J|Q|K|A|
| Display |2|3|4|5|6|7|8|9| 0|-|=|≡|H|
```

The following may more legible choice:
```
| Cards   |                          |
|---------|--------------------------|
| Value   |2|3|4|5|6|7|8|9|10|J|Q|K|A|
| Display |0|1|2|3|4|5|6|7|8 |9|-|=|≡|
```

### Future Ideas
- Add sounds
- Save/Display high score
- Add a "Win" animation
- Consider using lap indicator for larger score limit


joeycastillo:
Played with it a bit this morning — this is fun! Holding off on merging tho because I do think that your alternate choice might indeed be more legible. As it stands it is a little bit confusing that the value `0` is higher than `2` and I think that might throw people for a loop — but the use of graphical elements for the "face" cards of Q, K and A does indeed feel pretty intuitive to me. 

Open to your feedback, happy to merge whenever you respond one way or the other!

Chris-E-J-Ellis:
> Played with it a bit this morning — this is fun! Holding off on merging tho because I do think that your alternate choice might indeed be more legible. As it stands it is a little bit confusing that the value `0` is higher than `2` and I think that might throw people for a loop — but the use of graphical elements for the "face" cards of Q, K and A does indeed feel pretty intuitive to me.
> 
> Open to your feedback, happy to merge whenever you respond one way or the other!

Thanks for taking a look, and also for creating such a great platform! The alternative card faces only really occurred to me as I was writing up the instructions, and I think you're probably right they seem to require a touch less brainpower to figure out. Appreciate the second opinion!

I'll update the code either later today or tomorrow and re-merge in `main`, thanks!

movement/watch_faces/complication/higher_lower_game_face.c
```diff
@@ -0,0 +1,357 @@
...
+} game_state_t;
+
+static game_state_t game_state = HL_GS_TITLE_SCREEN;
+static card_t game_board[GAME_BOARD_SIZE] = {0};
+static uint8_t guess_position = 0;
+static uint8_t score = 0;
+static uint8_t completed_board_count = 0;
+
+static uint8_t generate_random_number(uint8_t num_values) {
+    // Emulator: use rand. Hardware: use arc4random.
```
matheusmoreira, line 78: Why the difference?

matheusmoreira (approved):

movement/watch_faces/complication/higher_lower_game_face.c
Chris-E-J-Ellis: I largely cribbed what some other faces did, but my understanding that is the emulator doesn't support the `arc4random` function, which I believe less predictable RNG.

matheusmoreira (approved):

matheusmoreira:

Issue #261, opened 2023-08-11
Bug in `watch_utility_date_time_to_unix_time` for leap years


rieck:
There seems to be a bug in the conversion between the `watch_date_time` struct and unix timestamps for leap years. The function `watch_utility_date_time_to_unix_time` in `watch_utility.c`  incorrectly adds an extra day when converting to a Unix timestamp. 

In my understanding, the dates and the timestamps should be identical, as we are just moving back and forth between representations. Maybe I am missing something here. I plan to look into the respective conversion functions later today but wanted to report the issue already now.

Best regards & thanks for this awesome project, 
Konrad

PR #262, opened 2023-08-11
Fix for incorrect conversion from `watch_date_time` to Unix time.

rieck:
This pull request aims at fixing an issue with time conversion. In my understanding, the current implementation of `watch_utility_date_time_to_unix_time` is incorrect. The function converts a `watch_date_time` struct to a Unix timestamp. However, It does not return the correct timestamp for leap years. I couldn't figure out exactly where the incorrect calculation occurs, but the following test cases help illustrate the problem:

- Leap year: `2020-01-01 00:00:00 GMT`. This date should should be converted to `1577836800`.
- Non-leap year: `2021-01-01 00:00:00 GMT`.  This date should should be converted to `1609459200`.

The following code displays both values in the simulator.

```c
    watch_date_time wd;
    wd.reg = 0;

    wd.unit.year = 0; // leap year
    wd.unit.day = wd.unit.month = 1;
    printf("t=%lu\n", watch_utility_date_time_to_unix_time(wd, 0));

    wd.unit.year = 1; // non-leap year
    printf("t=%lu\n", watch_utility_date_time_to_unix_time(wd, 0));    
```

In the current implementation the first test case returns `1577923200` (incorrect) and the second `1609459200` (correct). 

Instead of reverse engineering the current implementation of `watch_utility_convert_to_unix_time`, I adapted the corresponding code from musl libc, which has already been used in other parts of the project. The code is surprisingly complex and I suspect that the old version misses some edge cases. With this pull request, both Unix times are now calculated correctly.

I am a little nervous about this fix, as it affects a large number of watch faces (16 complications and 3 clocks). If this is a misunderstanding on my part, I apologize for the inconvenience and the false alarm.

Best regards, Konrad

### DIFF
--- a/watch-library/shared/watch/watch_utility.c
+++ b/watch-library/shared/watch/watch_utility.c
@@ -102,13 +102,81 @@ uint16_t watch_utility_days_since_new_year(uint16_t year, uint8_t month, uint8_t
     return (is_leap(year) && (month > 2) ? 1 : 0) + DAYS_SO_FAR[month - 1] + day;
 }
 
-uint32_t watch_utility_convert_to_unix_time(uint16_t year, uint8_t month, uint8_t day, uint8_t hour, uint8_t minute, uint8_t second, uint32_t utc_offset) {
-    uint32_t year_adj = year + 4800;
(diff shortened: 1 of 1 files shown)


rieck:
The changes aim at fixing #261.

WesleyAC (approved):

PR #268, opened 2023-09-02
Add a manual DST toggle to settings

harmonicoscillator:
Hi! I'm excited about receiving a sensorwatch lite soon, but noticed that there was no DST handling currently available in the settings. 

My hope is that this makes it easier to keep faces like the sunrise/sunset accurate by updating both the hour and the timezone simultaneously, so you don't e.g. forget to update the timezone while editing the hour, leading to sunrises off by one hour.

## Contents

This PR adds a very simplistic DST manual toggle by:
- Adds a new binary toggle to `movement_settings_t` , `active_dst`, to track whether DST is active or not
- Adds two 'jump tables' that defines how time zones should be shifted when going to and from active DST
  - Using the currently defined 40 timezones there is no way to make the two jump tables 1-1, so I took some liberties and optimized for what I think are the most common scenarios (US and European timezones)
  - This means there are cases where if you set your timezone to a few that definitely don't observe DST, you'll get some odd behavior toggling DST multiple times 
- Added a page to the `set_time_face` to toggle DST
- Toggling the DST setting updates the hour and timezone simultaneously

## Potential side effects

Short of using the real `tzinfo` db, I don't think there's any sane way to automatically update the world clock settings, so I simply chose not to.

This addition shouldn't affect anyone who chooses not to use it (lucky folks without DST) other than an additional page in `set_time_face`. 

## Testing

I tested this out in the wasm emulator and things seemed to work OK. 

## Potential future work

I'd be interested in writing a future PR that adds datetime settings for when to automatically turn DST on and off, potentially using the RTC alarm callbacks.

Is there any chance at all that the actual [tzinfo full db](https://www.iana.org/time-zones) would fit in the firmware? If it's possible, then using linux-style timekeeping where the hardware RTC is always set to UTC and all time displays go through the tzinfo conversion would greatly simplify things like the world clock (where, since different regions do DST at different dates, it's pretty manual to know "what's the time in London?" when my local time is US EDT.)


matheusmoreira:
> Is there any chance at all that the actual [tzinfo full db](https://www.iana.org/time-zones) would fit in the firmware?

That'd be great indeed... The website says the full data distribution is about 440 KB but the microcontroller only has a maximum of 256 KB of flash. Perhaps there's a way to include only the data that the user is interested in.

harmonicoscillator:
I've been running this since last October on my actual watch (red board) and used it successfully for the latest "spring forward" without any issues. Any hope this can get into mainline?

>> Is there any chance at all that the actual [tzinfo full db](https://www.iana.org/time-zones) would fit in the firmware?

> That'd be great indeed... The website says the full data distribution is about 440 KB but the microcontroller only has a maximum of 256 KB of flash. Perhaps there's a way to include only the data that the user is interested in.

I did a little bit of digging and there are projects that compress the tzinfo db to be much smaller, generally by throwing away historical information. I couple examples I found were:
- https://github.com/turbosree/TinyTZ
- https://github.com/evq/utz (this one claims to fit in ~3kB)
If one of those were incorporated into the SensorWatch build system I _think_ we could get best-guess zone handling for future times. They're both quite old project though, so would likely need some TLC.

matheusmoreira (approved):

voloved:
Auto-DST PR created that branched off this PR.
https://github.com/joeycastillo/Sensor-Watch/pull/470

drtrigon:
> I've been running this since last October on my actual watch (red board) and used it successfully for the latest "spring forward" without any issues. Any hope this can get into mainline?

I built a firmware using the (inofficial) online tool and this is missing. When can I expect it to be released there? I would love to get it too.

harmonicoscillator:
@drtrigon this PR was merged [but then reverted](https://github.com/joeycastillo/Sensor-Watch/commit/fe9a0a693d4a9f721c589dad5d7499c408aea198), so it's not in mainline either.

joeycastillo:
It was merged but then reverted in favor of a more robust DST solution which is now in [Second Movement](https://github.com/joeycastillo/second-movement), the 2.0 refactor.

Second Movement is almost ready for general use, and handles DST changeovers both in the main clock and in all world clock time zones.

drtrigon:
Thanks for the answer! Looking forward to use that feature.

PR #276, opened 2023-09-16
Simulator volume slider

hchargois:
The buzzer in the simulator plays a triangle wave at full gain, which is _extremely_ loud and annoying, at least to my ears.

This PR adds a volume slider to control the buzzer's volume. The value is saved to the browser's local storage, so it stays the same between reloads of the page.

While I was at it, I added the selected skin in local storage too.

### DIFF
--- a/watch-library/simulator/shell.html
+++ b/watch-library/simulator/shell.html
@@ -882,10 +882,22 @@
       </g>
     </g>
   </svg>
-  <table cellpadding="5"><tr><td id="skinselect">
-          <input type="radio" id="f91w" name="skin" value="f91w" onclick="toggleSkin()" checked><label for="f91w">F-91W</label>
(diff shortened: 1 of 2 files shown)


WesleyAC (approved):
This makes the skin switching buttons a little awkward — I think just hiding the radiobuttons and only showing the labels would help, and adding a bit of whitespace in between them. But, overall, looks good — this is something that's bugged me as well.
screenshot

hchargois:
Ah, I didn't see this before as it only shows up on tiny screens.

I took that opportunity to improve the page design a bit.

Here's a before/after comparison on a phone screen.

screenshot_simulator_before

screenshot_simulator_after

WesleyAC:
Looks great, thanks!

PR #278, opened 2023-09-17
Bug: Event race condition

hchargois:
I've noticed that button presses are sometimes "skipped", i.e. I press a button and no `BUTTON_UP` event arrives to the face's loop function. I've also noticed that the higher the requested `tick_frequency`, the more often buttons presses are skipped.

I've created this PR to demonstrate the issue, with a simple face that runs at a 64 Hz tick frequency, and counts up on each press of the alarm button, or at least it should. When running in the simulator, for 10 presses of the button, the counter only goes up by around 3 to 5, so more than half of the button presses are skipped.

What I _think_ happens, is that:
* the button press makes the button interrupt handler and callback run
* the global `event.event_type` is set to the correct `BUTTON` event
* but, the face's loop function is not called immediately, and before it is called...
* a tick happens, the tick handler sets the `event.event_type` to `TICK`, "overwriting" the `BUTTON` event into oblivion
* finally the face's loop is called with the `TICK` event

I've added a naive fix in which the tick handler avoids overwriting the event if it's not empty. This seems to fix the counter face: 10 button presses now reliably increment the counter 10 times.

I don't expect this fix to be the best way to fix the issue, so I don't expect this PR to be merged, I just wanted to demonstrate the bug and give weight to the race condition theory above with a simple fix.

I imagine a proper fix may be to add a stack of pending events and call the face's loop multiple times in a row if there are multiple pending events. Maybe the tick events also need to be handled differently than the other ones: if we are falling behind tick events, we don't want to replay them all, we usually want only the "latest" tick event.

I should also say that I don't have the actual board yet, so these observations are made with the simulator. But since I think this bug is in `movement.c` and not in the watch library, I expect the same bug to be exhibited by the hardware board. Unless there's something different with how interrupts are handled.

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -632,7 +632,6 @@ void cb_fast_tick(void) {
 }
 
 void cb_tick(void) {
-    event.event_type = EVENT_TICK;
     watch_date_time date_time = watch_rtc_get_date_time();
     if (date_time.unit.second != movement_state.last_second) {
         // TODO: can we consolidate these two ticks?
@@ -644,4 +643,8 @@ void cb_tick(void) {
     } else {
         movement_state.subsecond++;
     }
+    if (event.event_type != EVENT_NONE) {
+        return;
+    }
+    event.event_type = EVENT_TICK;
 }
--- a/movement/make/Makefile
+++ b/movement/make/Makefile
@@ -118,6 +118,7 @@ SRCS += \
   ../watch_faces/complication/flashlight_face.c \
   ../watch_faces/clock/decimal_time_face.c \
   ../watch_faces/clock/wyoscan_face.c \
+  ../watch_faces/demo/event_race_face.c \
 # New watch faces go above this line.
 
 # Leave this line at the bottom of the file; it has all the targets for making your project.
--- a/movement/movement_faces.h
+++ b/movement/movement_faces.h
@@ -95,6 +95,7 @@
 #include "flashlight_face.h"
 #include "decimal_time_face.h"
 #include "wyoscan_face.h"
+#include "event_race_face.h"
 // New includes go above this line.
 
 #endif // MOVEMENT_FACES_H_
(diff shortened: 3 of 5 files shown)


matheusmoreira:
I have also noticed this problem on hardware but it's especially noticeable in the emulator.

> I imagine a proper fix may be to add a queue of pending events and call the face's loop multiple times in a row if there are multiple pending events. Maybe the tick events also need to be handled differently than the other ones: if we are falling behind tick events, we don't want to replay them all, we usually want only the "latest" tick event.

Completely agree. A circular buffer seems like a good way to implement this queue.

Buffering events will add latency which could impact user experience. The target is a microcontroller running on bare metal: can we make any guarantees regarding event handling times?

movement/movement.c
```diff
@@ -644,4 +643,8 @@ void cb_tick(void) {
     } else {
         movement_state.subsecond++;
     }
+    if (event.event_type != EVENT_NONE) {
+        return;
+    }
+    event.event_type = EVENT_TICK;
```
matheusmoreira, line 649: The PR should contain these lines. The watch face that reproduces this issue is of great interest to developers but unlikely to be interesting to users. Perhaps it should be a separate PR or branch.

Is there a way to update `event` atomically?

matheusmoreira (changes requested):
This fix will impact watch faces because it violates their assumptions. Many just display data on tick but some use the tick event to keep time by setting the tick frequency explicitly and counting the ticks. The pulsometer, for example, does this:

https://github.com/joeycastillo/Sensor-Watch/blob/af18673e1aa53091880d829a6fa4d7e23a6b4381/movement/watch_faces/complication/pulsometer_face.c#L92

These faces rely on the ticks having that exact frequency in order to keep time correctly. Dropping tick events would break that assumption by potentially making the tick events non-periodic.

In order for a proper fix to be implemented, delta time is needed. Movement would have to keep the time between each tick so that it can tell the watch face how much time has passed between the last tick event and the current one. The face logic would then be able to do any integration math it needs to figure out what changed based on the time difference. This way it becomes possible to work correctly with just the latest tick event.

In other words, movement would have to work just like a game engine. I haven't studied how feasible that is yet but I think such a thing should be perfectly possible. The watch face programming paradigm would probably change though.

Issue #275, opened 2023-09-15
hourly chime not working in standby mode


Niehztog:
Some time ago I flashed a new firmware to my sensor watch built directly from the up-to-date main branch. Since then the hourly chime is not working in standby mode. My watches buzzer is working fine, so it is not related to that. It must have been some of the commits from the last few months which broke the feature.

Can anyone running a more resent built of movement on his watch confirm this?

CarpeNoctem:
Hmm. I'm running a pretty fresh build. I'll enable the chime and comment back here later today.

CarpeNoctem:
I can confirm that my watch is missing these hourly chimes in LE mode as well. I'll check my other watch when I get home. Its firmware was compiled (by me) from source as well, but sometime earlier. If it doesn't have the issue, that could help narrow down when the issue crept in.

CarpeNoctem:
So I haven't made it home yet, but did another check out of curiosity. I set an alarm for a time during LE mode. Paying closer attention the second time, I confirmed that it actually wakes the watch up out of LE mode into normal mode. I'm not sure whether or not this used to be the case for the hourly chime, but may be worth noting.

fedpost:
I have had this issue for a while. It has been intermittent and I haven't cared enough to investigate or give it any further thought, but I can confirm it happens with my build as well.

Niehztog:
Commit c1580b356d16cc49db873b020d72465a5f8b3a45 replaced the function call that plays the hourly chime from [watch_buzzer_play_note](https://github.com/joeycastillo/Sensor-Watch/blob/main/watch-library/hardware/watch/watch_buzzer.c#L171-L180) with [watch_buzzer_play_sequence](https://github.com/joeycastillo/Sensor-Watch/blob/main/watch-library/hardware/watch/watch_buzzer.c#L75-L91).
`watch_buzzer_play_note` works in low energy mode, `watch_buzzer_play_sequence` obviously doesn't. (I tested that on my watch by reverting the change and the hourly chime worked again in low energy mode).

@TheOnePerson and @neutralinsomniac , does maybe one of you have an idea how we can get the beautiful custom hourly chime tunes from #209 working in low energy mode?

TheOnePerson:
As far as I know does neither of `watch_buzzer_play_note` or `watch_buzzer_play_sequence` work in LE mode. 
Just from looking at the code: 

[...]

PR #279, opened 2023-09-18
whiterose: add new clock face

jdek:
A simple sub-hour audible signal useful for tracking time without having to actually look at the watch itself.

Inspired by the TV show Mr. Robot:

> Each beep indicates one minute of my time that has passed.
> I have allotted you no more than three minutes.
> -- Whiterose

https://www.youtube.com/watch?v=GHI1Rq9oTh8 (spoiler)

### DIFF
--- a/movement/watch_faces/clock/clock_face.c
+++ b/movement/watch_faces/clock/clock_face.c
@@ -46,6 +46,20 @@
 #define CLOCK_FACE_24H_ONLY 0
 #endif
 
+/**
+ * Signal interval in minutes. Default is an hourly signal. Acceptable range for
+ * this value is 0 to 59. This can be defined at build time using:
+ *
+ *   CFLAGS="-DCLOCK_FACE_SIGNAL_INTERVAL=1" make
+ *
+ * For an interval of 1 minute. Keep in mind that this is a sub-hour interval,
+ * if the hour is not divisible by the value set then it may not work as
+ * expected.
+ */
+#ifndef CLOCK_FACE_SIGNAL_INTERVAL
+#define CLOCK_FACE_SIGNAL_INTERVAL 0
+#endif
+
 typedef struct {
     struct {
         watch_date_time previous;
@@ -286,6 +300,9 @@ bool clock_face_wants_background_task(movement_settings_t *settings, void *conte
     if (!state->time_signal_enabled) return false;
 
     watch_date_time date_time = watch_rtc_get_date_time();
-
+#if CLOCK_FACE_SIGNAL_INTERVAL == 0
     return date_time.unit.minute == 0;
+#else
+    return date_time.unit.minute % CLOCK_FACE_SIGNAL_INTERVAL == 0;
+#endif
 }


movement/watch_faces/clock/simple_clock_face.c
```diff
@@ -161,11 +184,37 @@ void simple_clock_face_resign(movement_settings_t *settings, void *context) {
...
 
     watch_date_time date_time = watch_rtc_get_date_time();
 
-    return date_time.unit.minute == 0;
+    if (date_time.unit.minute == 0)
+        return true;
+
+    switch (settings->bit.signal_freq) {
+    case SIGNAL_1MIN:
+            return date_time.unit.second == 0;
```
TheOnePerson, line 197: No need to check for `date_time.unit.second == 0` here and in the other cases. The `..._wants_background_task`-handlers are only called once a minute. Otherwise line the former line 170 would not have worked. I suggest you remove these checks, to keep things simple and streamlined.

movement/watch_faces/clock/simple_clock_face.c
```diff
@@ -134,16 +142,31 @@ bool simple_clock_face_loop(movement_event_t event, movement_settings_t *setting
...
+#if SIGNAL_SUB_SIMPLE_TUNE
+                if (date_time.unit.minute == 0)
+                    movement_play_signal();
+                else
+                    watch_buzzer_play_sequence(signal_sub_tune, NULL);
+#else
                 movement_play_signal();
+#endif
                 // and then turn the buzzer peripheral off again.
                 watch_disable_buzzer();
```
TheOnePerson, line 171: This will be problematic in conjunction with `watch_buzzer_play_sequence()`. See [this discussion.](https://github.com/joeycastillo/Sensor-Watch/issues/275#issuecomment-1730809624)

TheOnePerson (commented):
I like this PR very much! Here are my 5 cents...

joeycastillo:
While I like this as a concept, I'm hesitant to add it to the core of Movement just because it is a bit of an edge case, and it feels like a lot to ship this feature to everyone (including the additional support burden of explaining another settings screen). 

Instead I think you could accomplish this with a watch face, using the background task request to make the signals and the watch face itself as the configuration interface. It would be a way to make the feature opt-in for folks that wanted it, without going too deep into Movement's core code.

WesleyAC:
@jdek I believe @joeycastillo's suggestion was to make a new face that isn't a clock, but rather just has a setting for whether the beep is enabled and how frequently it happens, letting the user use whatever clock face they want. The beep will still happen, even if the face is in the background. Removing the clock from this, and adding a option to disable the beep altogether would get it there, I think.

matheusmoreira:
@jdek

> could it be integrated into the new `clock_face.c`

Absolutely! Would you like to do the honors? Something like this should work:

```C

/*
 * Sound the time signal every CHIME_PERIOD minutes.
 * The defualt is 0 for an hourly chime.
 * Set to 15 for quarter hourly signals or 1 for whiterose mode!
 */
#ifndef CLOCK_FACE_CHIME_PERIOD
#define CLOCK_FACE_CHIME_PERIOD 0
#endif

bool clock_face_wants_background_task(movement_settings_t *settings, void *context) {
    // ...

    if (CLOCK_FACE_CHIME_PERIOD == 0) {
        return date_time.unit.minute == 0;
    } else {
        return date_time.unit.minute % CLOCK_FACE_CHIME_PERIOD == 0;
    }
}
```

matheusmoreira:
> Honestly, I'm fine for it to go in as unconfigurable until you rework the preferences

> Will update the PR later today with an `#ifdef`'d & rebased version. Thanks.

Great! I'm preparing my `next` branch with the next batch of pull requests right now. I'll merge this feature in and test it on my own watch.

I encourage you to test my branch as well, that way you can test your proposed feature and several others in one go. It helps ensure everything is stable before merging. I'll make sure you are fully credited for your effort!

> I have two sensor watches, so I'll just carry one which each interval I use

Yeah... That you're resorting to carrying two watches is the best possible evidence that we need runtime settings for this. Due to work and life, I can't give you a timeframe guarantee on how fast I can get this done. Hopefully soon but I can't make promises.

> I don't like the idea of having preferences outside of the preferences face, it's messy and bad UI.

I wouldn't say "bad" as it's clearly been working... I'd say it's a little inconsistent. Most developers are individually defining the user interface for each face. There's variations in how they approach the problem. This variation can be confusing for the user. Centralizing it in the movement framework would normalize all that.

> Setting an interval should be done from a preferences face, turning it on and off could be done from the face itself.

I agree with this. I believe this the ability to toggle _some_ settings from within the watch face should be maintained as it is useful, fast and improves usability. Other settings which are more complex and less frequently modified however should probably be in a dedicated preferences face with a polished and consistent user interface.

In any case, to do this I'll be adding quite a bit of state to movement. I need to consider the impact this will have on RAM and ROM usage more carefully. I'll probably need to figure out a way to make all this wiring up as compile time static as possible.

PR #282, opened 2023-09-20
Implementation of an animated clock face.

rieck:
This watch face is an extension of the simple watch face and contains exactly the same functions. However, depending on a configurable interval, the watch face displays short animations to amaze your viewers and demonstrate the capabilities of the sensor watch.

Currently, the animated watch face supports two animations that are displayed at an interval of 1, 5, 10, 15, 30 or 60 minutes. The animations are:

1. **scroll clock**. With this animation, the current time is shown as text on the display, such as "ten twenty-two" for "10:22". Initially, I experimented with longer text but found it difficult to read due to the limitations of the segment display.

https://github.com/joeycastillo/Sensor-Watch/assets/81579/e9d8b8c2-33f4-40a6-953f-8b4632c77b45

2. **wyoscan clock**. This animation uses the amazing wyoscan clock by Willy Hardy. The animation runs three times, where it first covers the displayed time, then moves over it, and finally restores it.

https://github.com/joeycastillo/Sensor-Watch/assets/81579/016e00d3-55df-438e-bf53-d5e6e499dccc

A *long press* of the *light button* activates the settings of this clock face. Currently, you can only select the interval for the animations or disable them. I would have liked to use a different button here, but the others are already occupied by the simple clock.

I tried to construct the animated clock face in a generic way. The main structure contains a general-purpose buffer that can be used by different animations to store temporary data. For example, the buffer contains the scrolling text for the scroll clock and the segment configuration of the wyoscan clock.

### DIFF
--- a/movement/lib/animations/animations.c
+++ b/movement/lib/animations/animations.c
@@ -1,0 +1,307 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2023 Konrad Rieck
+ * Copyright (c) 2023 Willy Hardy
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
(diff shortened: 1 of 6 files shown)


wryun:
Neat! If you're interesting in messing about with animations like this, you might also be interested in the morph I did in https://github.com/joeycastillo/Sensor-Watch/pull/102

(it would be possible to morph every clock transition)

wjhrdy:
Love this!

rieck:
The current code is unstabled on the real watch. The reason is unclear. I will start from scratch.

PR #283, opened 2023-09-22
fix signal tunes not firing in background, and move buzzer disable/enable logic into movement

neutralinsomniac:


joeycastillo:
Are you seeing signal tunes firing in the background with this PR? I just tested and it didn't fire (using either the custom tunes or the default); the issue seems to me to be that `movement_play_tune` doesn't work at all in standby.

I see two options for this: 

1. in the simple clock face, if we see `SIGNAL_TUNE_DEFAULT` defined, use the legacy `movement_play_signal` function, which works in standby mode. This doesn't solve custom tunes, but it seems like the quickest way to get this feature restored for folks using the default builds. 
2. Fix `movement_play_tune` by having it set the TCC to run in standby mode when the tune begins, and then setting it to not run in standby when the tune is complete. This feels like it would require extra testing; I never anticipated running the buzzer in standby, so I don't know if it will "just work". It's also very important to make sure that the TCC is always stopped at the end of the tune, because the TCC keeps the main 4 MHz clock awake, which will significantly impact battery life if it's not disabled. 

I'm inclined to implement option A in this PR, and then leave the question of background tunes to a separate issue.

814d3:
I can confirm, that the default hourly signal is not working, when the watch is in low energy mode. I am using tha latest ([67be6af](https://github.com/joeycastillo/Sensor-Watch/commit/67be6affd3367c4f6c302a57fdb35fc2bda50ce6)) version. Only two short *clicks* are recognizable.
I did not test timers or the alarm sound, when in LE mode, but probably they won't work either.
Greetings

814d3:
I forgot to mention, that I built with `SIGNAL_TUNE_DEFAULT`, which should work with this merge, as I understand correctly?
@joeycastillo, should I create a new issue to point this out? This merge request can be overseen easily.

PR #289, opened 2023-09-28
Make tarot face more configurable

WesleyAC:
This largely adds variables and uses existing variables in places they should already be, but it also changes the default minimum number of cards to one, down from three. Since that's personal preference, feel free to drop that commit if you'd like.


movement/watch_faces/complication/tarot_face.c
```diff
@@ -32,6 +32,23 @@
 #include <string.h>
 #include "tarot_face.h"
 
+#define DEFAULT_MAJOR_ARCANA_ONLY true
+#define SHOW_REVERSED_CARDS true
+#define TAROT_MIN_CARDS_TO_DRAW 1
+#define TAROT_FACE_CARD_REMINDER false
```
neutralinsomniac, line 38: I say remove this configuration and just always show the rank

movement/watch_faces/complication/tarot_face.c
```diff
@@ -32,6 +32,23 @@
 #include <string.h>
 #include "tarot_face.h"
 
+#define DEFAULT_MAJOR_ARCANA_ONLY true
+#define SHOW_REVERSED_CARDS true
+#define TAROT_MIN_CARDS_TO_DRAW 1
```
neutralinsomniac, line 37: I like the default of 3 since it covers a "normal" spread but also can be used for a 1 or 2-card spread without changing this value (since you can just ignore the 2nd/3rd cards depending on your desired spread)

movement/watch_faces/complication/tarot_face.c
```diff
@@ -103,12 +120,22 @@ static void tarot_display(tarot_state_t *state) {
...
     // show a special status if we're looking at the first or last card in the spread
-    if (state->current_card == 0) {
-        start_end_string = "St";
-    } else if (state->current_card == state->num_cards_to_draw - 1) {
-        start_end_string = "En";
-    } else {
-        start_end_string = "  ";
+    start_end_string = "  ";
+    watch_clear_indicator(WATCH_INDICATOR_LAP);
+    if (TAROT_LOOP_STYLE != TAROT_LOOP_STYLE_NONE && state->num_cards_to_draw > 1) {
```
neutralinsomniac, line 125: if you want to keep the ability to configure this at compile time, change these `if`'s to `#ifdef` statements to avoid unnecessary code making its way into the compiled object.

PR #293, opened 2023-09-30
movement: Add "instant" led_duration option.

WesleyAC:
This illuminates the LED only for the time that the button is pressed, and turns it off as soon as it's released, which is the behaviour of the original watch.

I chose a led_duration of zero to represent "instant" and all bits set to represent "no LED", which is arbitrary but seemed more sensible to me.

This PR also makes this the default. This is a matter of personal preference, but "instant" is the behaviour of the original watch, and seems like the thing more people would expect. Feel free not to take that commit if you disagree, though.

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -213,12 +213,22 @@ void movement_request_tick_frequency(uint8_t freq) {
 }
 
 void movement_illuminate_led(void) {
-    if (movement_state.settings.bit.led_duration) {
+    if (movement_state.settings.bit.led_duration != 0b111) {
         watch_set_led_color(movement_state.settings.bit.led_red_color ? (0xF | movement_state.settings.bit.led_red_color << 4) : 0,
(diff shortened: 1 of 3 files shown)


WesleyAC:
Not tested on the physical watch yet, just the emulator. Will try it on the physical watch whenever I bother to take it apart and reflash it next.

Here's what the settings screen looks like:

WesleyAC:
Tested on hardware and seems to be working well

814d3:
Very nice, I will test it and report back after 1 to 2 weeks.

chris-v8:
My solution suggestion would be to merge both instant, and timed LED. In simple face, only the instant LED would be active. In every other face, LED could be 5, 3, n seconds. This would solve the conflict of of faces using short and long press. Although I wonder if making additional if statement would have a drawback on the CPU?

wryun:
To take the LED in the other direction (more light!), I raised https://github.com/joeycastillo/Sensor-Watch/pull/394/files

:)

matheusmoreira (approved):
Looks great.

PR #294, opened 2023-10-01
Use slower "fast ticks" (8 Hz instead of 128 Hz) to detect button press and light duration

alesgenova:
Every time a button is pressed, the light turns on, or a sound is playing, Movement enters a 128 Hz fast tick mode.

The `cb_fast_tick()` function is called 128 times per second. In this function we do among other things some checks on the time a button is held down or the amount of time left on an alarm playback.

With this PR I propose to lower the frequency of the fast tick mode to 8 Hz (the frequency is adjustable by changing `FAST_TICK_FREQUENCY`). This way we will reduce the invocation of  `cb_fast_tick()` by over an order of magnitude.

We potentially lose a little bit of accuracy (up to 0.125s at 8 Hz) in the amount of time needed for a button longpress (0.5s +/- 0.125), or the light duration (1s +/- 0.125), but I think this inaccuracies are not noticeable.

Another consequence of this PR is that the 128 Hz mode is no longer reserved by Movement, and any face can now request this tick frequency.

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -22,7 +22,8 @@
  * SOFTWARE.
  */
 
-#define MOVEMENT_LONG_PRESS_TICKS 64
+#define FAST_TICK_FREQUENCY 8
+#define MOVEMENT_LONG_PRESS_TICKS FAST_TICK_FREQUENCY / 2
 
 #include <stdio.h>
 #include <string.h>
@@ -134,6 +135,7 @@ void cb_alarm_btn_extwake(void);
 void cb_alarm_fired(void);
 void cb_fast_tick(void);
 void cb_tick(void);
+void cb_fast_and_regular_tick(void);
 
 static inline void _movement_reset_inactivity_countdown(void) {
     movement_state.le_mode_ticks = movement_le_inactivity_deadlines[movement_state.settings.bit.le_interval];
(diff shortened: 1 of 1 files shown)


wryun:
:+1: seems like a good thing to me.

wryun:
I've been using this on my watch without issue.

matheusmoreira (commented):
Do any of the maintainers have any concerns over an 8 Hz fast tick as opposed to 128 Hz?

The code itself looks good to me. @wryun has tested this on hardware so this PR can be tagged `tested-on-hw` and he can be credited on the merge commit with a git trailer:

    Tested-by: James Haggerty (@wryun) ...

Can also be used to credit reviewers and maintainers who signed off on it:

    Reviewed-by: Matheus ...
    Reviewed-by: Other reviewers ...
    Signed-off-by: Maintainer ...

wryun:
As a continuing data point, this has been fine for me for 6 months.

matheusmoreira:
@voloved Do you think this will impact debouncing in any way?

voloved:
> @voloved Do you think this will impact debouncing in any way?

It would. This would now make the lowest reliable debounce time be a quarter second, which is too high for debounce. Though there are other ways to debounce, such as creating a new timer each time the button is pressed and killing it once the debounce time is over.

Using 8 Hz also causes any watch face that uses 8Hz to also need to be modified to not use 8Hz. Though they can be moved to 16Hz and checked if their subec mod 8 is zero.

What is the benefit of making this change to begin with? Does it save battery? Is the battery savings noticeable on the watch's life?

alesgenova:
@matheusmoreira @voloved 
There is logic in this PR for faces to be able to use the same frequency as the `FAST_TICK_FREQUENCY`.
This effectively removes the limitation that 128Hz (or 8Hz now) was reserved and not available to faces.

I don't know the impact to battery life, as I don't have the equipment to test it.

Issue #291, opened 2023-09-28
Allow easier configuration of compile-time settings


WesleyAC:
Currently, compile-time changes seem to be done by either changing the code, or with `#define`s in individual faces. This means that making changes to constants requires doing a git merge when updating the firmware. Instead, I'd like to see a file just for user preferences that could be kept outside of version control for changing constants. (or perhaps these could be set in the `movement_config.h`, since that's currently where I would expect most users to start their customization).

I would imagine this looking like a extremely simple macro or two, maybe as simple as a `DEFINE_CONSTANT(name, default)` that is just a `#ifndef` / `#define` / `#endif`, where the user's configuration file is loaded before watch faces. I think most of the value is just in having a standard way of doing this across faces, the implementation is pretty trivial.

Thoughts? Perhaps related to #70, although that is somewhat different and fancier, I think probably we want to solve this issue and that one separately.

PR #295, opened 2023-10-02
add compile-time options for all preferences to movement_config

madhogs:
PR to add compile-time options for all preferences to movement_config.h.
I have set the values to match all currently existing defaults.

This will allow the user to omit the preferences face if they wish and define all options at build time.
Addresses issue - https://github.com/joeycastillo/Sensor-Watch/issues/291 (some of it at least)


matheusmoreira (approved):
This is great! More compile time customization features is always good.

Further improvements can be built on top of this in the future: setting flags like these could be made to cause the settings variables to be deleted and their references turned into compile time constants. This will allow compilers to partially evaluate functions and to eliminate dead code automatically, making for a smaller and more efficient program.

matheusmoreira:
Am now running this code on real hardware. Used it to set 24h as default. Worked. No issues.

madhogs:
@matheusmoreira That's great to hear, thanks for looking at this. Let me know if there is anything more needed to merge this in.

matheusmoreira:
> Let me know if there is anything more needed to merge this in.

I think the upstream maintainers are reviewing it right now. I'm looking forward to it.

In any case, it's already been merged into my branch[¹] along with several other pull requests. I'm running that branch on my watch with no issues so far. You're welcome to use it if you'd like!

[¹]: https://github.com/matheusmoreira/sensor-watch/tree/advanced

814d3:
@madhogs, since there are "instant" LED activation and 10 min. LE timeout, there need to be some additions in your code, right? I'm not so familiar with this, so just a note.

madhogs:
@814d3 yes, this was merged after the 10min timeout option was merged and unfortunately needs changing. I have adjusted the options in my next pr here https://github.com/joeycastillo/Sensor-Watch/pull/387 . This fixes the comment and puts the default back to 1 hour.
The instant option hasn't been merged yet (i'm pretty sure) so it should be correct in my latest PR. If that does get merged first i will update my pr.

PR #299, opened 2023-10-07
Support leading zero representation for 24h clock

glines:
This adds a mode to display 24h clock with leading zero rather than with the 24h indicator. For instance, "6:00 24h" appears as "06:00" with this mode enabled. (Finally! After all these years, I no longer need to squint to check if my watch says "24h" or if I accidentally bumped the alarm button and it says "PM"!)

The new mode is "hidden" behind the existing 24h option. Switch between leading-zero mode and the default with long-press of alarm button when viewing the screen with 24h setting. With leading-zero mode enabled, the options screen displays "024h". Otherwise, the existing 24h option works as before, so the change should not be disruptive to existing users.

This change occupies an additional global preferences bit, but I'm hoping this feature is popular enough to justify the cost. (If some day it does not fit in the BKUP register and needs to be stored elsewhere, I don't mind.)

Impact to performance should be minimal, as the bit only needs to be checked once every hour.

The only thing I'm not super happy about is all of the additional format strings. If there's a way to manually force the display to show a leading zero without using a format string, please let me know and I will rewrite this patch.

### DIFF
--- a/movement/movement.h
+++ b/movement/movement.h
@@ -60,9 +60,10 @@ typedef union {
         // time-oriented complication like a sunrise/sunset timer, and a simple locale preference could tell an
         // altimeter to display feet or meters as easily as it tells a thermometer to display degrees in F or C.
         bool clock_mode_24h : 1;            // indicates whether clock should use 12 or 24 hour mode.
+        bool clock_24h_leading_zero : 1;    // indicates whether clock should leading zero to indicate 24 hour mode.
         bool use_imperial_units : 1;        // indicates whether to use metric units (the default) or imperial.
         bool alarm_enabled : 1;             // indicates whether there is at least one alarm enabled.
-        uint8_t reserved : 6;               // room for more preferences if needed.
+        uint8_t reserved : 5;               // room for more preferences if needed.
     } bit;
     uint32_t reg;
 } movement_settings_t;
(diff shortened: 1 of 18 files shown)


glines:
I revised this patch to call `watch_display_string("0", 4)` after updating the display, rather than using multiple nearly identical format strings. This approach is much easier to follow and maintain IMO.

I expanded this patch to add leading zero support to any watch face I could find that reads the `clock_mode_24h` bit. I don't use all of these watch faces myself, so my testing is limited on some of them. I made sure they at least compile and display 12h and 24h time properly in the simulator.

glines:
Fixed small bug, 24h indicator was not shown in the alarm face. (this is a bug in the current implementation, not a regression from this patch)

movement/movement.h
```diff
@@ -60,9 +60,10 @@ typedef union {
         // time-oriented complication like a sunrise/sunset timer, and a simple locale preference could tell an
         // altimeter to display feet or meters as easily as it tells a thermometer to display degrees in F or C.
         bool clock_mode_24h : 1;            // indicates whether clock should use 12 or 24 hour mode.
+        bool clock_24h_leading_zero : 1;    // indicates whether clock should leading zero to indicate 24 hour mode.
```
matheusmoreira, line 63: A maintainer should clarify the project's policy on ABI stability before this can be reviewed. This will shift the meaning of all following bits which unfortunately breaks the binary interface for anyone saving and loading these settings to storage as a binary structure. There seems to be at least two open pull requests implementing this feature right now so I'd say there are people out there depending on such interfaces.

matheusmoreira (commented):
The code looks good, just need clarification on the breaking settings structure change.

movement/movement.h
matheusmoreira: Looks like movement 2.0 will store settings in LFS. Might even have settings for each watch face. This doesn't seem to be much of an issue anymore.

PR #305, opened 2023-10-20
Add Couch-To-5k training face

ekaitz-zarraga:
This is a watch face that follows this program for running:

http://www.c25k.com/

The program is designed to train 3 times a week. Each training is a *session*. Each of the rounds you have in the training is an *exercise*.
 
The training goes like this:  5min warm-up walk -> Run X minutes -> Walk Y minutes -> ... -> Stop
 
The watch face shows it like this:
The weekday indicator shows if you need to Warm Up (`WU`), run (`rU`), walk (`WA`) or stop (`--`).
 
The month-day indicator shows the session you are in (from 1 to 27).
 
The timer shows the time you have left in the exercise and the exercise you are doing (MM:SS:ee). When an exercise finishes you are notified with an alarm. When the whole session finishes, a different tone is played for a longer period.
 
Pressing the ALARM button pauses/resumes the clock.
 
Pressing the LIGHT button does nothing if the timer is not paused. When it is paused it clears the current session (it restarts it to the beginning) and if it was already cleared or the current session was finished moves to the next session.


ekaitz-zarraga:
About the changes proposed:

- I'm not sure about a couple of details. I don't know if repainting the whole screen each time is ok or if it's too much for energy consumption. I considered other ways to do it using `watch_display_character` but it's a private function so I guess I wasn't supposed to use it.

- I also wrote all the training in a large structure in memory, maybe it's too much and it's better to do it using `uint8_t`s that represent 10s units. That would reduce the memory footprint to the half, but would need to compute a little and make the program slightly harder to read.

There are other things I don't have very clear, but I'm open for suggestions.

Thanks! It was a lot of fun to do!

WesleyAC (changes requested):
Looks good overall! The main thing I'd ask is that you rename it `couch_to_5k_face` to fit in with the naming convention of the other faces. You should also consider if you want to do anything for entering LE mode — I think returning to the default face would be reasonable, since even on the lowest timeout, the exercise will be completed before we enter LE mode.

In terms of energy usage, it's slightly better to just update the characters that changed, but it's not a big deal, and I think it's not worth it for this face, which people would only be using once a day for around half an hour.

Saving some space would be good, but I also feel fine about merging this without that. I think the easiest way to do that would be to make `C25K_SESSIONS` a double-pointer, so you don't need to repeat the duplicate sessions in memory.

ekaitz-zarraga:
> Looks good overall! The main thing I'd ask is that you rename it `couch_to_5k_face` to fit in with the naming convention of the other faces.

Yeah, that looks better. Done. 30c8da7f553146e22be456f0852351dff7269e1f

> You should also consider if you want to do anything for entering LE mode — I think returning to the default face would be reasonable, since even on the lowest timeout, the exercise will be completed before we enter LE mode.

Good, I added this too. f77ddc83b8b1ac5b9516496d0f9ead7865026ce5

> In terms of energy usage, it's slightly better to just update the characters that changed, but it's not a big deal, and I think it's not worth it for this face, which people would only be using once a day for around half an hour.

Good!

> Saving some space would be good, but I also feel fine about merging this without that. I think the easiest way to do that would be to make `C25K_SESSIONS` a double-pointer, so you don't need to repeat the duplicate sessions in memory.

It is already an array of pointers that points to the arrays of the session values. During the first weeks the session is repeated 3 times per week, and it's using the same array for them. There are other ways to reduce memory usage, but I also agree this is ok.

Thanks for the comments and suggestions.

WesleyAC (approved):
Looks good! And yeah, I wasn't reading carefully when looking at `C25K_SESSIONS` — i wouldn't worry about that, it's only 180 bytes out of 240KB we have to work with, which is 0.07%.

PR #310, opened 2023-10-30
Wyoscan update

wjhrdy:
This update brings 2 enhancements:
1. no more low power timeout.
2. respects the 12-hour and 24-hour setting, even though that was not an option on the original watch.


WesleyAC (approved):
Looks good!

One thing that's a little bit odd about this is that the LE timer doesn't get reset until `_movement_reset_inactivity_countdown` is called, and that doesn't happen correctly after you change modes away from wyoscan, you need to press another button after that — but that's a movement bug, not a bug in this PR.

WesleyAC (commented):
Actually, looking more at this, it's not a simple fix to get the `le_interval` to update properly, so I'd like @joeycastillo's thoughts on how this should work before merging this.

mcguirepr89:
Hello -- thank you all for this wonderful project. I wanted to see if there is any update on getting these changes pulled into `main`?

PR #314, opened 2023-11-03
Replace secondary face index with "pages" system.

WesleyAC:
This keeps the exact same behaviour in the default case, but allows for having more than two pages of faces, useful for navigating a large number of faces.

The first commit implements this change.

The second commit adds a variable for configuring which faces are hidden, instead of hiding all but the first page of faces.

The third commit allows more configuration of mode long-press behaviour.

It's very possible that there's a nicer way to expose/configure this, I understand that this might be providing more knobs than people want. I'm happy to try to implement other ideas or write more thorough comments if that would help.

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -249,13 +249,13 @@ void movement_move_to_face(uint8_t watch_face_index) {
 }
 
 void movement_move_to_next_face(void) {
-    for (uint32_t i = 0; i < MOVEMENT_NUM_PAGES; i++) {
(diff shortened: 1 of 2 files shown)


WesleyAC:
Trying this on my watch and I'm not convinced it's the right system. I have some other ideas that I'm going to try out for making navigating through many faces easier, but I'm going to close this for now. Would be curious to hear thoughts from anyone who gives it a shot, though!

PR #315, opened 2023-11-03
Add solstice_face

WesleyAC:
A face for telling the dates and times of solstices and equinoxes

It shows the upcoming solstice or equinox by default. The upper right number
is the year, and the bottom numbers are the date in MMDD format. Use the
alarm / light buttons to go forwards / backwards in time. Long press the
alarm button to show the time of the event, including what weekday it is on,
in your local timezone (DST is not handled).

Supports the years 2020 - 2083. The calculations are reasonably accurate for
years between 1000 and 3000, but limitations in the sensor watch libraries
(which could easily be worked around) prevent making use of that.


WesleyAC:
Going to go ahead and merge all my own new faces, since they've been sitting for a while and any bugs in them won't affect anyone who doesn't choose to install them.

PR #316, opened 2023-11-03
moon_phase_face: Make alarm long-press reset to current day.

WesleyAC:


matheusmoreira (approved):
Code seems correct to me. This is a much needed feature. I tried to do this when I got my watch and was confused when it didn't work.

matheusmoreira:
Am now running this code on real hardware. Works well, no issues.

PR #321, opened 2023-11-09
Add page_face

WesleyAC:
This allows for organization of a large number of faces, by hiding some faces in a "page"

This seems like a better path forward for what I was trying to do with #314, although it's worth thinking at some point about how this might interact with something like #264 (since it currently wouldn't play nicely with that sort of system).


matheusmoreira (commented):
This is a feature I really want too, and this face is a really creative use of the control over scrolling. Looks good to me.

Before I found this PR, I thought about implementing it in a different way in the movement framework itself by adding a layer of indirection: the watch faces array would become an array of pages which is an array of watch faces, short pressing mode scrolls within a page while long pressing mode scrolls the pages. What do you think of this approach?

Seems you had a similar idea with PR #314. Why were you convinced that was not the right approach? It requires less memory than multiple `page_face` instances. Could be made to work with #264 too.

chris-v8:
any chance this could make it to main before movement overhaul? I've been running this PR for few months now on my watch with zero issues. Works really good for hiding settings/preferences.

matheusmoreira:
@chris-v8 The feature freeze for the refactoring is due in September. I can't promise it'll make it in but I'll try. I need to review all the code and referenced PRs and see what we can do. It's also something that will significantly change the product's user experience so I cannot merge it without discussing that first.

I can tell you this is something I want personally. Right now movement has two hardcoded pages that we switch between by holding the button. I feel like it should be simple to generalize this mechanism so that we can have any number of pages.

In any case, thanks for testing it on real hardware, it's greatly appreciated and if it does get merged I'll make sure everyone's names are in the commit.

PR #322, opened 2023-11-10
Add save_load_face

WesleyAC:
This is essentially a less-invasive version of #85, which operates entirely as a watch face, allowing users to save preferences to LFS and load them again after reflashing (or at any time, really). It saves all eight backup registers (although currently only three are actually used — movement settings, location, and "birthday") and the RTC time, although the RTC time is only used in the UI and not loaded.


WesleyAC:
Going to go ahead and merge all my own new faces, since they've been sitting for a while and any bugs in them won't affect anyone who doesn't choose to install them.

PR #323, opened 2023-11-11
i18n: Add compile-time language setting

WesleyAC:
This currently supports Spanish and German (with English being the default), and simply translates the weekdays, although more translations should be added.

German translations are from @Kaoala on GH Discussions. Spanish translations are from ltzamatema on Discord.

DO is a little hard to read (it looks a little like OO if you're not expecting it). Possibly dO would look better, but I've chosen to keep everything in caps for now. If someone who uses the watch in another language would like to change it, feel free.


xurizaemon:
Great to see this @WesleyAC! Thanks.

Here's a link to [my writeup of the Reo Māori translation](https://chris.bur.gs/sensor-watch-r%C4%81-m%C4%81ori/) - similarly to `DO`, I found `AP` wouldn't render on the day characters, and settled for now on `A ` - appreciate input from anyone with more than my limited knowledge.

I expect this would be used with `make LANGUAGE=SPANISH` when building. I looked for where we'd add documentation to clarify that for folks, seems like [this page](https://www.sensorwatch.net/docs/movement/building/#compile-firmware) touches on compile time flags, and it would be a separate MR for a different target branch.

WesleyAC:
I saw you mention those on Discord and just added them in a second commit :)

The reason `AP` can't display properly is that the two segements on the left side and the two segments on the right side of that character are each tied together — https://www.sensorwatch.net/docs/wig/display/#limitations-of-the-weekday-digits has a diagram that explains it. I think that just showing "A" is a good solution.

And yep, it's used in that way. I'll send a PR to update the docs as well.

xurizaemon:
Just want to say I'm using this and the PR has my thumbs up to accept :) thanks @WesleyAC

PR #324, opened 2023-11-12
Enable LTO

WesleyAC:
This reduces the size of the build by ~8% and may improve speed or reduce power usage as well.

Yet to be tested on hardware.


WesleyAC:
This seems to completely break. When I flash the generated firmware, the watch immediately goes back into bootloader mode. I think it would be worth debugging this, since the potential gains are pretty big, but I suspect it might be complicated, maybe something not declared as volatile that needs to be, or some other such nonsense.

Gonna close this for now.

PR #326, opened 2023-11-13
Enable LTO

WesleyAC:
This saves around 9.5% of space on the default build, and likely reduces power consumption as well, although I'm not sure by how much. Numbers from TI [1] suggest that it might be a ~2-3% speed improvement (they measure 3.8% using -O3, but we're using -Os so we should expect slightly less), which would mean somewhere between 12 and 20 extra days of battery life.

The change to the .vectors section is from a blog post [2] that did a excellent job explaining the exact problem I had.

[1]: https://software-dl.ti.com/codegen/docs/articles/cgt_Build_a_Leaner_and_Faster_TI_Arm_Application_with_Link_Time_Optimization.html
[2]: https://m0agx.eu/fixing-cortex-m-startup-code-for-link-time-optimization.html


WesleyAC:
Putting the LTO flags in `make.mk` results in significantly more optimization (reducing space by 7.97% as opposed to 4.29%), but breaks the build for starter-project, which accidentally uses a bunch of stdlib stuff that doesn't actually exist, a latent bug that is triggered by trying to use LTO. The correct flags could likely fix this, things like `-lnosys`, `-specs=nano.specs`, and `-specs=nosys.specs`, but I couldn't manage to get this working, unfortunately. Will probably take another stab at it later.

matheusmoreira:
LTO and whole program optimization is vital for reducing code size!

> which accidentally uses a bunch of stdlib stuff that doesn't actually exist

Please elaborate, which missing symbols are referenced? I have some experience reducing `libc` footprint. Will try to build this when I have more time.

wryun:
I've been running this for a while. Seems good (as you would expect).

Issue #198, opened 2023-01-30
Index out of bounds error (SIGSEGV) in emulator when setting longitude on sunrise/sunset face


eshrh:
Steps to reproduce on my computer:

0. Clone
1. cd movement/make
2. emmake make
3. On the emulator:
	1. m m (switch to sunrise/sunset face)
	2. long press a (enter edit mode)
	3. l l l l l (advance to LO)
	4. l (advance to first digit)
	5. a (set hundreds to 1)
	6. l l l l l (advance to sunrise/sunset)

On firefox, the last l causes the light to stay on and throws an index out of bounds error somewhere in the wasm.
On chromium, the webpage crashes and shows SIGSEGV.

Any thoughts?

It doesn't seem to crash on every latitude entered, just some values every time.

I also have this error, not sure if it's relevant:
```
../../littlefs/lfs.c:1229:error: Corrupted dir pair at {0x0, 0x1}
Ignore that error! Formatting filesystem...
Filesystem mounted with 7680 bytes free.
```

WesleyAC:
I can reproduce this in the simulator (although notably it doesn't seem to affect the hardware, possibly the overflow is happening but not overwriting anything important)

PR #329, opened 2023-11-19
sunrise_sunset_face: Fix use of uninitialized memory.

WesleyAC:
This was causing a crash in the simulator when setting the location.

Fixes: #198

### DIFF
--- a/movement/watch_faces/complication/sunrise_sunset_face.c
+++ b/movement/watch_faces/complication/sunrise_sunset_face.c
@@ -197,6 +197,8 @@ static void _sunrise_sunset_face_update_settings_display(movement_event_t event,
     char buf[12];
 
     switch (state->page) {
+        case 0:
+            return;
         case 1:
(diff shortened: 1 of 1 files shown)


movement/watch_faces/complication/sunrise_sunset_face.c
```diff
@@ -197,6 +197,8 @@ static void _sunrise_sunset_face_update_settings_display(movement_event_t event,
     char buf[12];
 
     switch (state->page) {
+        case 0:
+            return;
```
matheusmoreira, line 201: Future improvement ideas: the page numbers with could be replaced with `enum` constants. The compiler can warn when `switch` statements don't cover all enumerated constants and also when they lack a `default` case.

matheusmoreira (approved):
The change fixes the undefined behavior.

matheusmoreira:
Am now running this code on real hardware. No issues.

PR #330, opened 2023-11-19
Fix simulator keyboard shortcuts

hchargois:
For context, currently in the simulator, A, L and M keys are shortcuts to "press" Alarm, Light and Mode buttons.

There was a few issues with that.

First issue is that these shortcuts prevented typing these letters in the debug console input.

Second thing is more a matter of taste; even though A, L and M are pretty logical, they are unfortunately physically located on a keyboard in the "wrong" places: A is on the left side of the keyboard while on the watch it's on the right, and for L it's the opposite. This made the shortcuts very counter-intuitive to actually use. For this reason, I added the arrow keys as shortcuts. Arrow up is the Light (the button on the top-left of the watch, arrow down (or left) is the Mode (the button on the bottom-left of the watch), and arrow right is the Alarm (button on the bottom-right).

Arrows have the benefit that contrary to any other letter keys, we can expect them to be placed in the correct disposition on all keyboards.

### DIFF
--- a/watch-library/simulator/shell.html
+++ b/watch-library/simulator/shell.html
@@ -907,13 +907,13 @@
     </div>
   </div>
 
-  <div style="display: flex; flex-direction: column; width: 100%">
+  <form onSubmit="sendText(); return false" style="display: flex; flex-direction: column; width: 100%">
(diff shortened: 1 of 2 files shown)


WesleyAC (approved):
Looks good, thanks!

PR #337, opened 2023-12-08
make the zero in wyoscan a little more visually appealing

theAlexes:
The Wyoscan app on the iPhone uses an asymmetric decay pattern for 0 instead of the oldest-first pattern that this replica face uses. This new ordering of 0 segments makes the decaying 0 less visually confusable with the decaying 1.


WesleyAC:
Looks good to me! If anyone is unhappy with this we can add a `#define` to change it back, but this seems uncontroversial.

PR #338, opened 2023-12-11
Add Butterfly game face

hchargois:
Butterfly is a 2-player game of shape recognition and speed/reflexes.

I believe it is the first multiplayer game for Sensor Watch, and it's not turn-based but real-time: both players play simultaneously.

It's a game of my own invention. The full rules are in the .h, but basically:

* player 1 plays with the LIGHT button, player 2 with ALARM
* the watch displays a symmetrical 2-character shape (a "butterfly") that both players need to memorize
* the watch then proceeds to show different butterflies, one per second
* when the target butterfly is shown, the fastest player to press their button wins the round
* if a player makes a mistake and presses their button when an incorrect butterfly is shown, they lose the round; and it's easy to make a mistake since butterflies are all very similar!

Screenshot_20231209-020503

(In this screenshot, player 1 has 5 points, player 2 has 2 points. It is the start of the round, the butterfly that they have to memorize is the "sixes" in the middle of the screen. When either player presses their button, the watch will start displaying different butterflies, and when the "sixes" butterfly is displayed again, the fastest player wins a point!)

I hope you find it fun!


joeycastillo:
What a delightful idea, thank you for your contribution!

PR #339, opened 2023-12-16
did a tiny bit of code cleanup, encountered a bug that might fix some sleep-mode issues

theAlexes:
while renaming the `current_watch_face` and `next_watch_face` to be clear that they are indexes into the `watch_face` array, we updated the app loop to use a pointer to the current face. This made us realize that the face loop can be visited twice in one turn, and thus cause conflict in `can_sleep`. This is resolved by boolean-ANDing the previous result in.


WesleyAC:
Looks good to me. Slightly dubious about the last commit in the stack but that seems like personal preference as much as anything, so I'm fine with merging it as-is.

PR #340, opened 2023-12-17
Work around various silicon errata

theAlexes:
Per the [Microchip docs](https://ww1.microchip.com/downloads/aemDocuments/documents/MCU32/ProductDocuments/Errata/SAM-L22-Family-Silicon-Errata-and-Data-Sheet-Clarification-DS80000782.pdf), one erratum may freeze the microcontroller (the SUPC/VREG operating in standby), another may cause increased power consumption from the TRNG not being properly disabled, and a third can hard-fault the chip if a standby occurs at the same time as a SysTick. This patch applies the recommended workarounds.

Also, we noticed that the sleep controller documentation specifies that you have to wait for the value to settle in the sleep mode register before actually `WFI`-ing, so we've modified the HAL to do that.

### DIFF
--- a/watch-library/hardware/watch/watch_private.c
+++ b/watch-library/hardware/watch/watch_private.c
@@ -110,18 +110,18 @@ int getentropy(void *buf, size_t buflen) {
         }
     }
 
-    watch_disable_TRNG(TRNG);
+    watch_disable_TRNG();
     hri_mclk_clear_APBCMASK_TRNG_bit(MCLK);
 
     return 0;
 }
 
-void watch_disable_TRNG(Trng *hw) {
-    hri_trng_clear_CTRLA_ENABLE_bit(hw);
-    // silicon erratum: the TRNG may leave internal components powered after disable.
-    // the workaround is to clear the register twice.
-    hri_trng_write_CTRLA_reg(hw, 0);
-    hri_trng_write_CTRLA_reg(hw, 0);
+void watch_disable_TRNG() {
+    // per Microchip datasheet clarification DS80000782,
+    // silicon erratum 1.16.1 indicates that the TRNG may leave internal components powered after being disabled.
+    // the workaround is to disable the TRNG by clearing the control register, twice.
+    hri_trng_write_CTRLA_reg(TRNG, 0);
+    hri_trng_write_CTRLA_reg(TRNG, 0);
 }
 
 
--- a/watch-library/hardware/watch/watch_private.c
+++ b/watch-library/hardware/watch/watch_private.c
@@ -35,6 +35,10 @@ void _watch_init(void) {
 
     // Use switching regulator for lower power consumption.
     SUPC->VREG.bit.SEL = 1;
+    // work around a silicon erratum that causes the microcontroller to lock up on leaving standby:
+    // request that the voltage regulator run in standby, and also that it switch to PL0.
+    SUPC->VREG.bit.RUNSTDBY = 1;
+    SUPC->VREG.bit.STDBYPL0 = 1;
     while(!SUPC->STATUS.bit.VREGRDY); // wait for voltage regulator to become ready
 
     // check the battery voltage...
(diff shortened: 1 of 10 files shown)


watch-library/hardware/watch/watch_private.c
```diff
@@ -106,12 +106,21 @@ int getentropy(void *buf, size_t buflen) {
...
     }
 
-    hri_trng_clear_CTRLA_ENABLE_bit(TRNG);
+    watch_disable_TRNG(TRNG);
     hri_mclk_clear_APBCMASK_TRNG_bit(MCLK);
 
     return 0;
 }
 
+void watch_disable_TRNG(Trng *hw) {
```
WesleyAC, line 119: Is there a reason to take `TRNG` as a arg here?

WesleyAC (changes requested):
Thanks for catching these!

Could you add references to specific section/page/etc numbers in the manual? That would be useful just to double check that this is all good.

I'm also curious if you have an idea of whether the `RUNSTDBY` / `STDBYPL0` will affect power usage, although if the docs suggest it we should presumably do it either way, especially given recent reports of freezing issues.

watch-library/hardware/watch/watch_private.c
theAlexes: I dunno, a lot of the ASF functions do the same thing, taking an unused argument and having the hardware register defined elsewhere. Will probably rewrite this in the process of annotating the errata.

theAlexes: we've rewritten this accordingly :)

theAlexes:
> Thanks for catching these!
> 
> Could you add references to specific section/page/etc numbers in the manual? That would be useful just to double check that this is all good.

updated accordingly, and included a link to the errata in the GH description.

> I'm also curious if you have an idea of whether the `RUNSTDBY` / `STDBYPL0` will affect power usage, although if the docs suggest it we should presumably do it either way, especially given recent reports of freezing issues.

power consumption is something we're pretty fuzzy on, unfortunately.

matheusmoreira (approved):
Looks great to me. Well referenced and explained by the comments, ensuring any future readers know why things work this way.

matheusmoreira:
> > I'm also curious if you have an idea of whether the `RUNSTDBY` / `STDBYPL0` will affect power usage, although if the docs suggest it we should presumably do it either way, especially given recent reports of freezing issues.
> 
> power consumption is something we're pretty fuzzy on, unfortunately.

That seems to be case for us all. There are some power consumption figures in the sensor watch documentation but they were measured with a [Power Profiler Kit II]. Absent access to such a profiler, I don't believe any objective statements regarding the power cost of any proposed change can be made.

In any case, if the silicon errata tells us to do something, we should probably do it regardless of any impact on power consumption, right?

[Power Profiler Kit II]: https://www.nordicsemi.com/Products/Development-hardware/Power-Profiler-Kit-2

matheusmoreira:
Am now running this code on the real hardware. No problems so far.

matheusmoreira:
My Sensor Watch Lite board is apparently unable to enter sleep mode when running this code. Have you or anyone else experienced this sleep mode issue?

 - Reproduced the issue with multiple low energy timer preferences set
   - 10 minutes
   - 1 hour
 - Confirmed that the clock face is returning `true` from its loop
   - Via simulator debugging
   - Indicates that it's not blocking the sleep mode
 - This is the only patch set in my branch which touches sleep mode

The most recent change in `main` to sleep mode functionality:

https://github.com/joeycastillo/Sensor-Watch/blob/af18673e1aa53091880d829a6fa4d7e23a6b4381/movement/movement.c#L529-L536

I suspect it could be somewhat related due to short circuiting behavior which is potentially unintended and might be a bug. This is what actually happens:

```C
// first trip  | can sleep | cannot sleep | can sleep    | cannot sleep
// second trip | can sleep | never runs   | cannot sleep | never runs
//          && | can sleep | cannot sleep | cannot sleep | cannot sleep
```

I need further testing to determine what impact this has, if any. However it appears you authored this change so I was hoping you'd clarify.

matheusmoreira:
> This is the only patch set in my branch which touches sleep mode

I stand corrected. It seems to be working perfectly. I eventually isolated the issue to the deadline watch face, PR #266.

PR #346, opened 2024-01-01
Expose character mapping internals and add invert

wryun:
Allowing the faces to access the internal segment mapping code makes it easier to have custom characters and have fancy effects.

This is the refactoring which was originally done for https://github.com/joeycastillo/Sensor-Watch/pull/102 with a few tweaks to keep it up to date.

This also adds the ability to invert the display (bonus).


matheusmoreira (changes requested):
Exposing the character segmap data is an excellent improvement which makes for more readable implementation!

I think the code for inverting the display belongs on a separate pull request so it can be evaluated independently from the aforementioned change. Notably, none of the watch faces updated by the commit make use of this new display inversion functionality.

wryun:
Heh, indeed it would make sense to split the invert. But the whole reason this code exists was to support the 'invert' (see linked PR), and I really wanted to sneak it in so I have less to update. I feel like the chances of getting the invert merged on its own are low!

It someone indicates they'd actually accept this if I removed the invert I'll do it. But I'd love to have the whole thing merged :)

PR #350, opened 2024-01-07
Handle visibility for tomato watchface

### LINKED ISSUE #349: Tomato watchface corrupt display when timer expires while in background
if a tomato timer expires while another watchface is shown it is going to update the display anyway, corrupting the display.

it seems that the tomato watchface update the display when a timer expires independently from it being visible or not.

pfmaggi:
Handle visibility for the tomato watchface so that a timer expiring while another watchface is shown does not corrupt the screen.

This fixes #349 (tested manually)


joeycastillo:
Good catch, thanks for this!

PR #269, opened 2023-09-05
French revolutionary face

CarpeNoctem:
Initial submission of French Revolutionary (Decimal) time face. ~~I've only run this on the emulator, but have just loaded it onto my watch for testing as well.~~ I'll try to add the power-saving stuff to the output in the coming week or so.

### French Revolutionary Decimal Time

Similar to the Decimal Time face, but with the day divided into ten hours instead of twenty four.
Each hour is divided into one hundred minutes, and those minutes are divided into 100 seconds.
I came across this one the Svalbard watch site here: https://svalbard.watch/pages/about_decimal_time.html
More info here as well: https://en.wikipedia.org/wiki/Decimal_time

### Usage

By default, the face just displays the current decimal time. Pressing the alarm button will toggle through other display options:
1) Just decimal time (with dT indicator at top)
2) Decimal time, with dT indicator and date above.
3) Decimal time, with 24-hr time above (where Day and Date would normally be displayed), BUT minutes first then hours.
   Sadly, the first character of the date area only goes up to 3. (see https://www.sensorwatch.net/docs/wig/display/#the-day-digits) I was going to begrudgindly leave this display option out when I realized that, but thought it would be better to have this backwards representation of the standard time than not at all.

A long-press of the light button will toggle the upper time between 12-hr AM/PM and 24-hr mode. I thought of reading the main setting for this, but thought that a person could normally prefer 12hr time, but next to a 10hr day want to see the normal time in the 24hr format.

A long-press of the alarm button will toggle the seconds off and on.

PR #360, opened 2024-01-17
template: fix compiler warning on watch_face_index as mentioned in PR269

CarpeNoctem:
Addresses a compiler warning of unused watch_face_index, as @WesleyAC mentioned in [a comment](https://github.com/joeycastillo/Sensor-Watch/pull/269#discussion_r1398458836) on Pull Request #269.
Putting the fix right at the root of where it came from - in the template.


WesleyAC:
Looks good, thanks!

PR #362, opened 2024-01-21
Revert "make the watch-face a global in movement.c, actually"

Kistelini:
This reverts commit 0e801ed505cb0c368bf1eb0473058efb6c275a3e.

This commit breaks the firmware on my watch (although it works in the simulator), the display doesn't show anything after reassembly. Maybe others can confirm this!


wryun:
It's declared as a const... this seems wrong. Presumably in the compiled version for the watch it's being put in a read-only place (but works in sim because no place is read only). Would have thought compiled would be complaining a lot.

Might be worth checking exactly what lead up to this. I'm also a bit dubious about 'wf' being pulled out at all, because theoretically someone could change the current face at various places in the app loop.

theAlexes:
ah yeah we forgot to edit the `const` out when moving that line up. im not sure how it worked on my watches.

theAlexes:
Anyway yeah, if changing watch faces inside the main loop is valid, then this change isn't, revert away 👍

PR #364, opened 2024-01-22
revert pointer nonsense, stick with array indexing

theAlexes:
the `const` qualifier on this is misplaced, so let's just go back to what we know works.

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -450,20 +450,17 @@ static void _sleep_mode_app_loop(void) {
 }
 
 bool app_loop(void) {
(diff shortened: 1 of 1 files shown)


theAlexes:
kinda torn on this change, actually, because at first glance, the array indexing generates slightly worse code.

taking off the `const` qualifier on the pointer version provokes warnings about discarding the qualifier, _and_ generates the same code as with `const` — probably it doesn't generate a read-only variable because it isn't in global scope, it's on the stack of `app_loop`.

so i think we can abandon this change for now, unless the pointer code actually causes issues with the index changing in the face code. (though isn't that guarded by a `watch_face_changed` flag in the movement state?)

PR #368, opened 2024-02-20
Added Dockerfile

rtfmkiesel:
Added a Dockerfile for building the firmware within a container. Also, opening up a PR on the docs for the instructions there.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM ubuntu:jammy-20240212
+
+RUN apt update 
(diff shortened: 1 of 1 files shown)


CarpeNoctem:
I'm not sure whether this is necessary; I wrote a Dockerfile like this when I got started as well, and was going to open a PR, but then I found there is already one in the project here:
[.devcontainer/Dockerfile](https://github.com/joeycastillo/Sensor-Watch/blob/main/.devcontainer/Dockerfile)
So I didn't bother adding mine.

PR #369, opened 2024-02-21
Structured TOTP watch face

matheusmoreira:
Improves the standard TOTP watch face by making it easier to define new secrets. Aggregating the TOTP data into a structure allows simple and functional definition of new instances with an initializer generator macro. This also makes the size and offset variables superfluous and so they are removed.


matheusmoreira:
Seems to be working well on real hardware. No issues so far.

PR #370, opened 2024-02-21
Forced 24h display build option

matheusmoreira:
Adds a build time configuration option which causes the entire movement framework to force a 24 hour clock display mode configuration. Applies to all watch faces which use the setting and resists attempts by watch faces to set a different mode. Updates the preferences watch face to hide the clock mode configuration page since there's no point in changing the setting.

All of these changes are conditionally compiled into the movement framework. The compiler is expected to eliminate all the forced 24h logic as dead code when not in this configuration.

It can be used via the build system:

    make FORCE_24H=yes

### DIFF
--- a/make.mk
+++ b/make.mk
@@ -229,3 +229,7 @@ endif
 ifeq ($(BOARD), OSO-FEAL-A1-00)
 CFLAGS += -DCRYSTALLESS
 endif
+
+ifdef FORCE_24H
+CFLAGS += -DMOVEMENT_FORCE_24H
+endif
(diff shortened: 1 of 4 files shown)


madhogs:
I've been using this on my watch for a bit now (with a small merge fix) and its working well. Like this PR as I never use the 12h option or even want it in the settings.

matheusmoreira:
@madhogs You've been using it? That's really cool!! It's a custom feature I developed for a customer. This was just my first attempt at it. In the end they preferred a smaller scale approach which is much closer to the 24 hour mode compile time default that was recently merged. I didn't think there would have been any interest in this PR and had all but abandoned it. Would you like me to rebase it?

PR #374, opened 2024-03-04
24h default option

matheusmoreira:
Allows the user to explicitly set the default clock mode to 24h display at compilation time with an option passed to the build system:

```shell
make MOVEMENT_CLOCK_MODE_DEFAULT_24H=yes
```

The current default behavior of the sensor watch is preserved.

As suggested by @joeycastillo in PR https://github.com/joeycastillo/Sensor-Watch/pull/373#pullrequestreview-1913150760.


matheusmoreira:
This feature has been merged into `main` by PR #380 and f35cb849de5ddd2abc6c1e242c0f8b3fe8049bd8, based on PR #295 which is a superset of this one. Thus there is no longer any need to merge this.

PR #382, opened 2024-03-14
WIP: High-precision timer

zzm634:
See https://github.com/joeycastillo/Sensor-Watch/discussions/381

WORK IN PROGRESS!

Still to do:
* Low level implementation using TC2/3
* Some control logic in movement.c
* Testing on hardware (especially edge cases)
* Simulator support
* Add countdown face as well for demo purposes

Not in scope for this PR:
* Refactor LEDs, buzzer, and long-press timing code to use HPT instead of fast-ticks mode

### DIFF
--- a/movement/make/Makefile
+++ b/movement/make/Makefile
@@ -126,6 +126,7 @@ SRCS += \
   ../watch_faces/clock/minute_repeater_decimal_face.c \
   ../watch_faces/complication/tuning_tones_face.c \
   ../watch_faces/complication/kitchen_conversions_face.c \
+  ../watch_faces/complication/hpt_lapsplit_chrono_face.c \
 # New watch faces go above this line.
 
 # Leave this line at the bottom of the file; it has all the targets for making your project.
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -58,6 +58,10 @@
 
 #include "movement_custom_signal_tunes.h"
 
+// macros that are not width-dependent
+#define SET_BIT(value, bit) value |= (1 << bit)
+#define CLR_BIT(value, bit) value = (~(~(value) | (1 << bit)))
+
 // Default to no secondary face behaviour.
 #ifndef MOVEMENT_SECONDARY_FACE_INDEX
 #define MOVEMENT_SECONDARY_FACE_INDEX 0
@@ -79,13 +83,18 @@ movement_state_t movement_state;
 void * watch_face_contexts[MOVEMENT_NUM_FACES];
 watch_date_time scheduled_tasks[MOVEMENT_NUM_FACES];
 
+// HPT stuff
+
 // bit flags indicating which watch face wants the HPT enabled.
-// make sure this number width is larger than the number of faces
+// TODO make sure this number width is larger than the number of faces
 #define HPT_REQUESTS_T uint16_t
 HPT_REQUESTS_T hpt_enable_requests;
 
-// timestamps at which watch faces should have a HPT event triggered. UINT32_MAX for no callback
-uint32_t hpt_scheduled_events[MOVEMENT_NUM_FACES];
+// timestamps at which watch faces should have a HPT event triggered. UINT64_MAX for no callback
+uint64_t hpt_scheduled_events[MOVEMENT_NUM_FACES];
+
+// the number of times the high-precision timer has overflowed
+uint8_t hpt_overflows = 0;
 
 const int32_t movement_le_inactivity_deadlines[8] = {INT_MAX, 600, 3600, 7200, 21600, 43200, 86400, 604800};
 const int16_t movement_timeout_inactivity_deadlines[4] = {60, 120, 300, 1800};
@@ -404,7 +413,7 @@ void app_setup(void) {
         for(uint8_t i = 0; i < MOVEMENT_NUM_FACES; i++) {
             watch_face_contexts[i] = NULL;
             scheduled_tasks[i].reg = 0;
-            hpt_scheduled_events[i] = UINT32_MAX;
+            hpt_scheduled_events[i] = UINT64_MAX;
         }
         hpt_enable_requests = 0;
         is_first_launch = false;
@@ -705,13 +714,37 @@ void cb_tick(void) {
     }
 }
 
+/** Figures out the next scheduled event for the HPT to wake up for.
+ * Must be called whenever:
+ * - a new event is scheduled
+ * - an old event is deleted
+ * - an event is triggered
+ * - the timer overflows
+ */
+static void _movement_hpt_schedule_next_event() {
+    uint64_t next = UINT64_MAX;
+    for(uint8_t req_idx = 0; req_idx < MOVEMENT_NUM_FACES; ++req_idx) {
+        uint64_t event = hpt_scheduled_events[req_idx];
+        if(event < next) {
+            event = next;
+        }
+    }
+
+    // if an event is scheduled and the timer is currently tracking this overflow cycle
+    if(next != UINT64_MAX && (next >> 32) == hpt_overflows) {
+        watch_hpt_schedule_callback((uint32_t)(next & UINT32_MAX));
+    } else {
+        watch_hpt_disable_scheduled_callback();
+    }
+}
+
 void movement_hpt_request(void) {
     movement_hpt_request_face(movement_state.current_face_idx);
 }
 void movement_hpt_request_face(uint8_t face_idx) {
     HPT_REQUESTS_T old_hpt_requests = hpt_enable_requests;
-    hpt_enable_requests |= (1 << face_idx);
-    if(old_hpt_requests != hpt_enable_requests) {
+    SET_BIT(hpt_enable_requests, face_idx);
+    if(old_hpt_requests == 0) {
         watch_hpt_enable();
     }
 }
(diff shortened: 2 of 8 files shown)


movement/movement.c
```diff
@@ -76,6 +82,20 @@
 movement_state_t movement_state;
 void * watch_face_contexts[MOVEMENT_NUM_FACES];
 watch_date_time scheduled_tasks[MOVEMENT_NUM_FACES];
+
+// HPT stuff
+
+// bit flags indicating which watch face wants the HPT enabled.
+// TODO make sure this number width is larger than the number of faces
+#define HPT_REQUESTS_T uint16_t
+HPT_REQUESTS_T hpt_enable_requests;
```
matheusmoreira, line 91: How about this?

```C
#define HPT_REQUESTS_SIZE (MOVEMENT_NUM_FACES / CHAR_BIT)

unsigned char hpt_enable_requests[HPT_REQUESTS_SIZE + 1];
```

movement/movement.c
```diff
@@ -76,6 +82,20 @@
...
+
+// timestamps at which watch faces should have a HPT event triggered. UINT64_MAX for no callback
+uint64_t hpt_scheduled_events[MOVEMENT_NUM_FACES];
+
+// the number of times the high-precision timer has overflowed
+uint8_t hpt_overflows = 0;
```
matheusmoreira, line 97: How many times are we expecting the timer to overflow?

movement/movement.c
```diff
@@ -76,6 +82,20 @@
...
+
+// HPT stuff
+
+// bit flags indicating which watch face wants the HPT enabled.
+// TODO make sure this number width is larger than the number of faces
+#define HPT_REQUESTS_T uint16_t
+HPT_REQUESTS_T hpt_enable_requests;
+
+// timestamps at which watch faces should have a HPT event triggered. UINT64_MAX for no callback
+uint64_t hpt_scheduled_events[MOVEMENT_NUM_FACES];
```
matheusmoreira, line 94: Perhaps it would be possible to store this in a watch face structure, eliminating a global variable.

movement/movement.c
```diff
@@ -691,3 +713,81 @@ void cb_tick(void) {
...
+ * - the timer overflows
+ */
+static void _movement_hpt_schedule_next_event() {
+    uint64_t next = UINT64_MAX;
+    for(uint8_t req_idx = 0; req_idx < MOVEMENT_NUM_FACES; ++req_idx) {
+        uint64_t event = hpt_scheduled_events[req_idx];
+        if(event < next) {
+            event = next;
+        }
+    }
```
matheusmoreira, line 731: It's critical that we make this function as efficient as possible. Hopefully we can eliminate this `O(n)` loop. What's your take on [radix timers]?

[radix timers]: https://github.com/romanf-dev/timers

movement/movement.c
```diff
@@ -691,3 +713,81 @@ void cb_tick(void) {
...
+    SET_BIT(hpt_enable_requests, face_idx);
+    if(old_hpt_requests == 0) {
+        watch_hpt_enable();
+    }
+}
+
+void movement_hpt_relenquish(void) {
+    movement_hpt_relenquish_face(movement_state.current_face_idx);
+}
+void movement_hpt_relenquish_face(uint8_t face_idx) {
```
matheusmoreira, line 755: Small correction: `relinquish`.

movement/movement.h
```diff
@@ -312,4 +313,85 @@ void movement_play_alarm_beeps(uint8_t rounds, BuzzerNote alarm_note);
...
+ * when your face no longer needs the timestamp provided by 
+ * "movement_hpt_get" or has no scheduled background tasks, in order to save
+ * power.
+*/
+void movement_hpt_relenquish(void);
+/**
+ * A variant of "movement_hpt_relenquish" that can be used when your face is
+ * not running in the foreground.
+*/
+void movement_hpt_relenquish_face(uint8_t face_idx);
```
matheusmoreira, line 372: Small correction: `relinquish`.

movement/movement.h
```diff
@@ -312,4 +313,85 @@ void movement_play_alarm_beeps(uint8_t rounds, BuzzerNote alarm_note);
...
+ * guarantee to be made about the HPT timestamp is that between the time your
+ * face calls "movement_hpt_request" until the moment it calls
+ * "movement_hpt_relenquish", the value returned from "movement_hpt_get" will
+ * increment upwards at 1024hz. Outside of that window, the timestamp value may
+ * change unpredictably.
+ * 
+ * Faces may schedule an EVENT_HPT event to occur by calling 
+ * "movement_hpt_schedule" and passing in a timestamp for the event to occur.
+ * The face must call "movement_hpt_request" before scheduling the event, and
+ * must not call "movement_hpt_relenquish" until after the event has occurred.
```
matheusmoreira, line 343: Why not? It won't be possible to cancel a running timer? What if the user cancels an alarm?

movement/movement.h
```diff
@@ -312,4 +313,85 @@ void movement_play_alarm_beeps(uint8_t rounds, BuzzerNote alarm_note);
...
+ * A variant of "movement_hpt_relenquish" that can be used when your face is
+ * not running in the foreground.
+*/
+void movement_hpt_relenquish_face(uint8_t face_idx);
+
+/**
+ * Schedules a future EVENT_HPT event to occur on or after the given timestamp.
+ * The Movement framework will do its best to trigger the event as close to the
+ * timestamp as possible, but it may be delayed if multiple faces or events are
+ * scheduled to occur on the same timestamp.
```
matheusmoreira, line 378: This implies timers are guaranteed to be handled, which means either real time code or a queue of pending timer events. Do you mean to guarantee this, and if so by which method?

I assume you meant real time constraints, as implied by your concerns in discussion #381:

> If interrupt handling takes too long, it's possible for the timer value to change in the background while we're handling scheduled events. If events are scheduled too close together or too close in the future, it might be possible that by the time we set the timestamp value for the new interrupt, the counter has already passed it and the interrupt won't trigger. Now, if the counter is running at 128hz and the cpu is running at 4mhz, that gives us about 31,000 clock cycles to finish up whatever we were doing and check the timer again. Plenty of time... unless a face calls _delay_ms() or has to wait to activate a peripheral. Not a deal-breaking problem, but something we need to be aware of.

However this comment was written before you updated your timer code to 1024 Hz. The deadlines will be much tighter.

watch-library/hardware/watch/watch_hpt.h
```diff
@@ -0,0 +1,68 @@
...
+#include <inttypes.h>
+
+/**
+ * Defines the reasons the HPT callback is being invoked. More than one flag may be set.
+*/
+typedef struct {
+    /**
+     * The callback is being invoked because of an error in the comparison (?)
+    */
+    bool cmp_error :1;
```
matheusmoreira, line 13: What conditions could possibly cause this?

matheusmoreira (commented):
Looking forward to seeing this develop.

movement/movement.c
zzm634: I don't think it's fair to call this O(n), because MOVEMENT_NUM_FACES is a constant. Most people are probably going to have like, 12 faces max on their watch, and even if they go hog wild, the most they can have is 256. Compilers are very good at optimizing simple loops like this, it will be fine.

radix timers seems like a really "clever" solution to this very simple problem.

movement/movement.c
zzm634: I thought I spelled that wrong. I'm gonna rename it to "release" anyway, because "relinquish" is too fancy.

movement/movement.h
zzm634: Oh I mean, they can call "relenquish" if they want to, but it means their scheduled event is going to be cancelled, and some devs might be surprised by that.

movement/movement.h
zzm634: I think the Movement framework needs to guarantee these events, but I haven't figured out exactly how to do it. It might be messy.

watch-library/hardware/watch/watch_hpt.h
zzm634: Not sure, but it was something the datasheet called out as a possible interrupt source. Unless I read the datasheet wrong, the timer only has a single interrupt vector and you're supposed to check the interrupt status register to see what caused it.

>Bit 1 – ERR: Error Interrupt Flag
This flag is set when a new capture occurs on a channel while the corresponding Match or Capture
Channel x interrupt flag is set, in which case there is nowhere to store the new capture.

It's entirely irrelevant to the HPT though, so maybe I should get rid of the flag in this struct.

movement/movement.c
zzm634: I had considered that, but either way, it's going to add up to one 64-bit timestamp per watch face. I didn't see an obvious watch face struct to add it to in movement.c, but I can't say I looked very hard.

movement/movement.c
zzm634: I only really need boolean flags, and I was trying to come up with a good way to pack them into a smaller structure to save memory.  (On the other hand, I went and allocated 64-bits for a bunch of clock face timestamps that probably aren't going to use them)

zzm634: Oh I see what you're doing with that. That would also work, and I think once I get this feature working, I'll do another pass and try to optimize memory usage as much as possible. (like, timestamps really only need to be 40-bits, even in 1024hz mode, but packing all that stuff up is a lot of extra math)

matheusmoreira: With some extra preprocessor hackery, we could get rid of that `+1` as well which is there just to prevent zero sized array errors in case the result is zero. That would save an extra byte.

watch-library/hardware/watch/watch_hpt.h
matheusmoreira: Match or capture? This has something to do with using these timers to generate waveforms. I don't think we'll be using that functionality. It could be safe to ignore that possibility.

movement/movement.h
matheusmoreira: We could check the nearest event to see if its timestamp is in the past. If so, that would mean its deadline was missed. We move forward in the event queue, processing them all in sequence until we find one with a timestamp in the future.

An efficient, hopefully constant time "next event" function would make it feasible. We'd be able to guarantee it would finish before the next timer window.

zzm634:
Just noticed that `watch_buzzer.c` uses TC3 for playing note sequences. Yet another thing for HPT to support I guess.

movement/make/Makefile
```diff
@@ -87,7 +87,6 @@ SRCS += \
   ../watch_faces/complication/ratemeter_face.c \
   ../watch_faces/complication/interval_face.c \
   ../watch_faces/complication/rpn_calculator_alt_face.c \
-  ../watch_faces/complication/stock_stopwatch_face.c \
```
zzm634, line 90: Before merging, I will refactor this face to use the HPT and add it back to the project.

movement/movement_config.h
zzm634, line 1: This will all be reverted when this is ready for merge.

movement/watch_faces/clock/hpt_led_test_face.c
zzm634, line 1: this face is likely to be deleted before merge as well

movement/watch_faces/complication/hpt_countdown_face.h
```diff
@@ -0,0 +1,104 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2024 <#author_name#>
```
zzm634, line 4: fix author name

watch-library/hardware/watch/watch_hpt.c
zzm634, line 1: needs copyright header

watch-library/shared/watch/watch_hpt.h
zzm634, line 1: copyright header

wryun:
Well, I'm a fan of higher preicions timing, and it all sounds good. But it's a little scary!

@joeycastillo - do you think this is a good approach?

joeycastillo:
I think I need to reserve an opinion on this until I can run power profiling on it, but I do like the idea of a shared high precision timer that everything that needs high precision timing can share, rather than having individual watch faces make use of TC peripherals. I’m also curious about the choice to use a 32 bit counter (which ties up two TC’s) instead of a 16 bit counter. Using a 16 bit counter would mean overflowing every minute or so, but it would make handling overflows as a matter of course instead of a thing that happens once in a blue moon. Especially since high precision timing mostly relates to things happening in the near future.  

@wryun your pinging me on this comes at an interesting moment too, because I’m currently doing a pretty big refactor of Movement in the background, tearing out a lot of low level technical debt and trying to free up resources for forthcoming hardware. In particular, my refactor should free up TC0 and TC1, which I think we might need to free up TC2 and TC3 for use by the accelerometer. Anyway I’m open to considering this for potential inclusion into the forthcoming second movement firmware, again, pending a more thorough power consumption analysis and a deeper review of the code.

wryun:
Thanks - happy to hear it's on your radar.

I guess the main advantage of the 32-bit counter from @zzm634 's perspective is that one might be able to do away with some of the overflow handling:

>   The code in this PR should handle this situation properly, but it does mean there is some extra logic in play to detect and compensate for timer overflows that occur while critical operations are being performed. These "safety" checks mean more CPU cycles, more power consumed, more latency for the user.

i.e. rather than having this complexity, one could just schedule an event before the overflow every time you start the timer like (EVENT_HPT_MAX_TIME or something) and document that code has to handle it by aborting the timer action. Since it's unlikely one needs a 49+ day precision timing event, and this might have an impact on battery life (and event hpt accuracy, in that you might be able to get the event to user code faster).

But given they're in there at the moment, I would also prefer a 16-bit counter and handle the overflow as a matter of course.

PR #383, opened 2024-03-16
uf2conv: argument to `re.split` should be a rawstring

theAlexes:
this should eliminate the invalid string escape warning seen in newer Pythons.


matheusmoreira (approved):
This is recommended by Python documentation.

PR #387, opened 2024-03-27
add the ability to set location and birthdate in movement config

madhogs:
Allow for location and birthdate preferences to be specified in the movement config. This means one less thing for the user to set up each time if they wish.
The defaults of 0 respect the current functionality, eg. for location the sunrise_sunset face still says 'no loc' and for birthdate the day_one face still defaults to 1959.

### DIFF
--- a/movement/movement_config.h
+++ b/movement/movement_config.h
@@ -76,15 +76,15 @@ const watch_face_t watch_faces[] = {
 /* Set the timeout before switching to low energy mode
  * Valid values are:
  * 0: Never
- * 1: 1 hour
- * 2: 2 hours
- * 3: 6 hours
- * 4: 12 hours
- * 5: 1 day
- * 6: 2 days
+ * 1: 10 mins
+ * 2: 1 hour
+ * 3: 2 hours
+ * 4: 6 hours
+ * 5: 12 hours
+ * 6: 1 day
  * 7: 7 days
  */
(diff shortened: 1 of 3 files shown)


movement/movement_config.h
```diff
@@ -76,15 +76,15 @@ const watch_face_t watch_faces[] = {
...
+ * 1: 10 mins
+ * 2: 1 hour
+ * 3: 2 hours
+ * 4: 6 hours
+ * 5: 12 hours
+ * 6: 1 day
  * 7: 7 days
  */
-#define MOVEMENT_DEFAULT_LOW_ENERGY_INTERVAL 1
+#define MOVEMENT_DEFAULT_LOW_ENERGY_INTERVAL 2
```
madhogs, line 87: This and the comment change above is setting the values back to their original defaults as there has been a change in the meantime adding the 10 minute option as value 1.

matheusmoreira (approved):
Great! I will be using this feature a lot.

madhogs:
I know this is my own PR but i've been running this on my physical watch for a few weeks now and all has been well and i've not seen any issues. I think this is ready to be merged without any issues or changes for anyone if any maintainers get a chance to look at this.

Merging this will also fix the defaults back to their original values (I can move that to a separate PR if wanted though)

Niehztog:
I have flashed this patch on both my watches and it works fine so far :+1:

PR #388, opened 2024-03-29
Fix compile errors and warnings in movement.c and shell.c

TheOnePerson:
Title says it all, really. Just a small PR to get rid of some compile errors and warnings.

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -319,11 +319,11 @@ void movement_request_wake() {
     _movement_reset_inactivity_countdown();
 }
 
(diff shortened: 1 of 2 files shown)


matheusmoreira (approved):
Could not reproduce the compilation problem but this still looks good. The more functions can be made static the better. Looks like an include was missing and it's fixed now.

wryun:
This is making it frustrating to try to update the builder. It would be great if someone could merge these fixes. @theAlexes @WesleyAC ?

PR #393, opened 2024-04-12
totp_face_lfs - reduce memory usage and add secmod

### LINKED ISSUE #392: Totp LFS face fails with certain codes and messes up the file system on physical watch
I was trying out the totp lfs face with the same codes as I use on the normal totp face but it not only fails with a watch reset but seems to cause some kind of filesystem breakage when it does.
Steps to recreate:
1. Flash board with current main branch, with addition of totp_face_lfs.
2. Open a serial connection (I used minicom) and run the following (these are example codes with the same format as my real ones):
```
echo otpauth://totp/Aaaaaaa?secret=JBSWY3DPEHPK3PXPJBSWY3DPEHPK3PXPJBSWY3DPEHPK3PXPJBAA&period=30&digits=6&issuer=AA&algorithm=SHA1 > totp_uris.txt
echo otpauth://totp/Bbbbb?secret=JBSWY3DPEHPK3PXPJBSWY3DPEHPK3PBB&period=30&digits=6&issuer=BB&algorithm=SHA1 >> totp_uris.txt
echo otpauth://totp/Ccccccc?secret=JBSWY3DPEHPK3PCC&period=30&digits=6&issuer=CC&algorithm=SHA1 >> totp_uris.txt
echo otpauth://totp/Ddddddd?secret=JBSWY3DPEHPK3PXPJBSWY3DPEHPK3PDD&period=30&digits=6&issuer=DD&algorithm=SHA1 >> totp_uris.txt
echo otpauth://totp/Eeeeeeeeeeeeeee?secret=JBSWY3DPEHPK3PXPJBSWY3DPEHPK3PXPJBSWY3DPEHPK3PXPJBSWY3DPEE&period=30&digits=6&issuer=EE&algorithm=SHA1 >> totp_uris.txt
echo otpauth://totp/Fffffffffffff?secret=JBSWY3DPEHPK3PXPJBSWY3DPFF&period=30&digits=6&issuer=FF&algorithm=SHA1 >> totp_uris.txt
echo otpauth://totp/Gggggggggggg?secret=JBSWY3DPEHPK3PGG&period=30&digits=6&issuer=GG&algorithm=SHA1 >> totp_uris.txt
echo otpauth://totp/Hhhhhhhhhhhh?secret=JBSWY3DPEHPK3PHH&period=30&digits=6&issuer=HH&algorithm=SHA1 >> totp_uris.txt
echo otpauth://totp/Iiiiiiiiiiiiii?secret=JBSWY3DPEHPK3PXPJBSWY3DPEHPK3PII&period=30&digits=6&issuer=II&algorithm=SHA1 >> totp_uris.txt
echo otpauth://totp/Jjjjjjjjjj?secret=JBSWY3DPEHPK3PXPJBSWY3DPEHPK3PJJ&period=30&digits=6&issuer=JJ&algorithm=SHA1 >> totp_uris.txt
echo otpauth://totp/Kkkkkkkkkkkk?secret=JBSWY3DPEHPK3PXPJBSWY3DPKK&period=30&digits=6&issuer=KK&algorithm=SHA1 >> totp_uris.txt
echo otpauth://totp/Llllll?secret=JBSWY3DPEHPK3PXPJBSWY3DPEHPK3PLL&period=30&digits=6&issuer=LL&algorithm=SHA1 >> totp_uris.txt
```
3. The above should have succeeded and you can see the above in the totp_uris.txt file.
4. Boot the physical watch and navigate to the totp lfs face.
5. When scrolling through the codes, it will loop after 4 (only AA <-> DD) in the above.
6. Navigating away from the lfs face now by pushing the mode button causes a hard reset of the watch.
7. If you now connect the board again to the computer via the usb port it will continuously flash red and the filesystem cannot be accessed (double clicking the button still works to flash it fortunately).

To fix the watch again after following the above I had to modify filesystem_init in filesystem.c to format the lfs storage on boot.

The above is all on the RED board.

Also just to clarify this is an issue on the totp_face_lfs only, the regular totp_face works fine with no issues.

wryun:
This should fix #392 

I've refactored this so it doesn't allocate the secrets on startup, but instead reads them from lfs each time. Not a fan of the filesystem API here... (would be so much more efficient if I had easy access to the underlying seek/read and could just leave the file open, but putting all that in this PR seemed like overkill).

I've also added a 'secure mode' which lets you remove from the secrets from the flash (see doco in header).

@madhogs - I haven't tried this on a real watch yet, so if you want to be a guinea pig that would be awesome :)

### DIFF
--- a/movement/watch_faces/complication/totp_face_lfs.c
+++ b/movement/watch_faces/complication/totp_face_lfs.c
@@ -35,29 +35,47 @@
 
 #include "totp_face_lfs.h"
 
-#define MAX_TOTP_RECORDS 20
-#define MAX_TOTP_SECRET_SIZE 48
+#define MAX_TOTP_RECORDS 30
+#define MAX_TOTP_SECRET_SIZE 128
 #define TOTP_FILE "totp_uris.txt"
 
 const char* TOTP_URI_START = "otpauth://totp/";
 
+/* The totp_record struct has two modes of operation:
+ *  - secmod = false: read from lfs (in which file_secret_offset
+ *    points to an offset in totp_uris.txt)
+ *  - secmod = true: read from memory
+ *    (in which case secret contains the decoded secret)
+ */
 struct totp_record {
-    uint8_t *secret;
-    size_t secret_size;
     char label[2];
-    uint32_t period;
     hmac_alg algorithm;
+    uint8_t period;
+    uint8_t secret_size;
+
+    union {
+        uint8_t *secret;
+        struct {
+            uint16_t file_secret_offset;
+            uint16_t file_secret_length;
+        };
+    };
 };
 
+/* This is used if we're not storing all the secrets but instead
+ * calculating them on demand. Avoids malloc in normal operation.
+ */
+static uint8_t current_secret[MAX_TOTP_SECRET_SIZE];
+
 static struct totp_record totp_records[MAX_TOTP_RECORDS];
-static int num_totp_records = 0;
+static uint8_t num_totp_records = 0;
 
 static void init_totp_record(struct totp_record *totp_record) {
-    totp_record->secret_size = 0;
     totp_record->label[0] = 'A';
     totp_record->label[1] = 'A';
-    totp_record->period = 30;
     totp_record->algorithm = SHA1;
+    totp_record->period = 30;
+    totp_record->secret_size = 0;
 }
 
 static bool totp_face_lfs_read_param(struct totp_record *totp_record, char *param, char *value) {
@@ -69,14 +87,13 @@ static bool totp_face_lfs_read_param(struct totp_record *totp_record, char *para
         totp_record->label[0] = value[0];
         totp_record->label[1] = value[1];
     } else if (!strcmp(param, "secret")) {
-        if (UNBASE32_LEN(strlen(value)) > MAX_TOTP_SECRET_SIZE) {
+        totp_record->file_secret_length = strlen(value);
+        if (UNBASE32_LEN(totp_record->file_secret_length) > MAX_TOTP_SECRET_SIZE) {
             printf("TOTP secret too long: %s\n", value);
             return false;
         }
-        totp_record->secret = malloc(UNBASE32_LEN(strlen(value)));
-        totp_record->secret_size = base32_decode((unsigned char *)value, totp_record->secret);
+        totp_record->secret_size = base32_decode((unsigned char *)value, current_secret);
         if (totp_record->secret_size == 0) {
-            free(totp_record->secret);
             printf("TOTP can't decode secret: %s\n", value);
             return false;
         }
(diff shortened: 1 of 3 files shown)


madhogs:
Thanks for looking into this :smile: 
Tested on my physical watch and the normal mode works great :partying_face:, tested with lots of very long 128 character secrets and again with my real secrets and everything worked well. Codes produced were correct and no memory issues, I also didn't see any filesystem issues at all with this code (previously i had to reformat it each time anything went wrong).

The 'secure mode' however did not work as expected, even with just 2 short example codes when activated it would work whilst remaining on that face however when moving away from the totp_lfs face and back again the watch would reset, which would then load 'no codes' as the file has been deleted.

I hope you don't mind me saying as it is a nice idea but considering the memory restrictions on the physical watch do you think it is worth keeping this 'secure mode'? Even in the best case I imagine it could only handle a few secrets before the watch would run out of memory.

movement/watch_faces/complication/totp_face_lfs.c
```diff
@@ -41,23 +41,41 @@
 
 const char* TOTP_URI_START = "otpauth://totp/";
 
+/* The totp_record struct has two modes of operation:
```
madhogs, line 44: Not code you changed but above here:
Is it worth increasing MAX_TOTP_SECRET_SIZE now?
This code worked with very long 128 character secrets but I had to adjust this variable to 128 to get it to work. 48 seems quite low.

Also I don't see any reason for MAX_TOTP_RECORDS to be as low as 20 either?

movement/movement_config.h
```diff
@@ -29,6 +29,7 @@
 
 const watch_face_t watch_faces[] = {
     simple_clock_face,
+    totp_face_lfs,
```
madhogs, line 32: I assume this is just for testing but this face should be removed from here before it is merged in to not change the defaults for anyone.

movement/watch_faces/complication/totp_face_lfs.c
wryun: Will do. The reason to have a max of 20 is that we're still allocating this array every time (and to me 20 seems a lot to scroll through), but I'll push it to 30.

madhogs:
I've been running the fix (I removed the secure mode code) on my physical watch for a week or so now and using the codes. All is working well, not had any issues :smile: 

In the interest of getting this merged would you be willing to separate the secure mode changes from the fix?
I would also be happy to do that with your name on the commit if you like.

wryun:
Feel free to raise PRs with this code with or without my name; at the moment I don't feel like splitting them ;)

Were you worried about having secure mode? It should be pretty hard to get into by accident.

madhogs:
I wasn't worried about activating it by accident, I just like to remove any features I'm not going to use on my own personal watch.
For the PR i just thought it makes sense to separate the fix from the new feature, can then merge the fix first whilst any issues in secure mode can be worked out :smile:

wryun:
Other PRs include this now.

PR #394, opened 2024-04-17
Keep light on if interacting

wryun:
This makes it possible to do a bunch of things without having to keep touching the light button.

I don't really see any downside with this. If you want the light to go off, just stop touching buttons.

WARNING: haven't yet tested on real hardware. Includes a minor patch to fix what I think is a bug in the simulator (without this, the watch will freeze up on transition as the buzzer causes the simulator to sleep in a state where the last animation frame was valid if the light was on).

### DIFF
--- a/watch-library/simulator/main.c
+++ b/watch-library/simulator/main.c
@@ -89,6 +89,7 @@ void main_loop_sleep(uint32_t ms) {
     main_loop_set_sleeping(true);
     emscripten_sleep(ms);
     main_loop_set_sleeping(false);
+    animation_frame_id = ANIMATION_FRAME_ID_INVALID;
 }
 
 bool main_loop_is_sleeping(void) {
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -543,6 +543,17 @@ bool app_loop(void) {
         event.subsecond = movement_state.subsecond;
         // the first trip through the loop overrides the can_sleep state
         can_sleep = wf->loop(event, &movement_state.settings, watch_face_contexts[movement_state.current_face_idx]);
+
+        // Keep light on if user is still interacting with the watch.
+        if (movement_state.light_ticks > 0) {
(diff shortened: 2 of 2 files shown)


814d3:
A nice feature, although I would at least recommend using the following mod, as it allows the LED to be used at about half the brightness level and at the same time provides better readability. This protects the LED (lifespan), the battery and your eyes ;-).
[etsy.com](https://www.etsy.com/de-en/listing/1448973768/back-light-spreader-for-casio-f-91w-a)

Kistelini:
I really like this feature, but I think it should be an option for the options menu because it changes "default" behavior of the watch.

wryun:
Hmm, surely by that logic everything is set in stone? (i.e. you can't change any of the default faces? Or is it just your concern that we're changing the button behaviour?)

watch-library/simulator/main.c
```diff
@@ -89,6 +89,7 @@ void main_loop_sleep(uint32_t ms) {
     main_loop_set_sleeping(true);
     emscripten_sleep(ms);
     main_loop_set_sleeping(false);
+    animation_frame_id = ANIMATION_FRAME_ID_INVALID;
```
matheusmoreira, line 92: This is a seemingly unrelated change -- what is the point of it?

matheusmoreira (approved):
I think this is a good feature.

watch-library/simulator/main.c
matheusmoreira: You're right, it's probably the simulator workaround. Now that you mentioned it, I re-read the pull request text and it makes sense.

I don't think this is causing any problems on real hardware. It's only compiled in simulator builds.

CarpeNoctem: Oh, you're right - I just saw that this change is inside the simulator folder. All good, then!

PR #408, opened 2024-06-26
Check that color is valid

tahnok:
Instead of merely checking that COLOR is set, check that it is one of RED, BLUE or GREEN

### DIFF
--- a/make.mk
+++ b/make.mk
@@ -213,6 +213,12 @@ ifndef COLOR
 $(error Set the COLOR variable to RED, BLUE, or GREEN depending on what board you have.)
 endif
 
+COLOR_VALID := $(filter $(COLOR),RED BLUE GREEN)
+
+ifeq ($(COLOR_VALID),)
+$(error COLOR must be RED, BLUE, or GREEN)
+endif
+
 ifeq ($(COLOR), BLUE)
 CFLAGS += -DWATCH_IS_BLUE_BOARD
 endif


theAlexes (approved):

matheusmoreira (approved):
Looks good to me, it will catch common user errors. It is a definite improvement!!

Some comments/ideas for possible future improvement:

The valid colors variable will need to be updated if a new board variant is released, this should probably be documented somewhere (I don't know where to be honest), perhaps also moved to a more noticeable place such as the top of the makefile, which would also allow it to abort earlier on user error.

The filter solution correctly identifies valid inputs but it'll still allow some invalid inputs:

```make
make COLOR='RED extra junk'
# Proceeds with COLOR='RED extra junk'
```

It happens because `RED` is treated as an independent word due to make's word splitting on spaces, and the filter function finds the RED word, despite the bad inputs. Users will have to try extra hard to break this but it's possible.

Inverting the logic with `filter-out` closes that loophole:

```make
COLOR_INVALID := $(filter-out RED BLUE GREEN,$(COLOR))

ifneq ($(COLOR_INVALID),)
$(error COLOR must be RED, BLUE, or GREEN)
endif
```

Whitelisting good input is generally better than blacklisting bad input.

This also has a problem with multiple valid inputs:

    make COLOR='RED GREEN'
    # Proceeds...

So we also need to count the words and ensure it's exactly 1:

```make
ifneq ($(words $(COLOR)),1)
$(error COLOR must be only one of RED, BLUE, or GREEN)
endif
```

I also suggest keeping the valid board colors in a variable for easy updating, eliminating the need to touch the logic and error messages in the future.

```make
COLOR.valid := RED BLUE GREEN
COLOR.error = $(error COLOR is [$(COLOR)] but it must be exactly one of these values: $(COLOR.valid))

ifneq ($(filter-out $(COLOR.valid),$(COLOR)),)
$(COLOR.error)
endif

ifneq ($(words $(COLOR)),1)
$(COLOR.error)
endif
```

PR #410, opened 2024-06-27
New watch face: Medical Thermometer

Kistelini:
This is heavily based on the temperature readout face. For further description see the header file.

### DIFF
--- a/movement/watch_faces/sensor/medical_thermometer_face.c
+++ b/movement/watch_faces/sensor/medical_thermometer_face.c
@@ -1,0 +1,151 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2024 Christian Buschau
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include <limits.h>
+#include <math.h>
+#include <stdlib.h>
+#include <string.h>
+#include "medical_thermometer_face.h"
+#include "thermistor_driver.h"
+
+static float _medical_thermometer_face_update(bool in_fahrenheit) {
+    thermistor_driver_enable();
+    float temperature_c = thermistor_driver_get_temperature();
+    char buf[14];
+    if (in_fahrenheit) {
+        sprintf(buf, "%4.1f#F", temperature_c * 1.8 + 32.0);
+    } else {
+        sprintf(buf, "%4.1f#C", temperature_c);
+    }
+    watch_display_string(buf, 4);
+    thermistor_driver_disable();
+    return temperature_c;
+}
+
+static void _medical_thermometer_face_clear(int last[]) {
+    for (size_t i = 0; i < LAST_SIZE; i++) {
+        last[i] = INT_MIN;
+    }
+}
+
+void medical_thermometer_face_setup(movement_settings_t *settings, uint8_t watch_face_index, void ** context_ptr) {
+    (void) settings;
+    (void) watch_face_index;
+    if (*context_ptr == NULL) {
+        *context_ptr = malloc(sizeof(medical_thermometer_state_t));
+        memset(*context_ptr, 0, sizeof(medical_thermometer_state_t));
+    }
+}
+
+void medical_thermometer_face_activate(movement_settings_t *settings, void *context) {
+    (void) settings;
+    medical_thermometer_state_t *state = (medical_thermometer_state_t *)context;
+    state->mode = MODE_NORMAL;
+    _medical_thermometer_face_clear(state->last);
+    watch_display_string("TE", 0);
+}
+
+bool medical_thermometer_face_loop(movement_event_t event, movement_settings_t *settings, void *context) {
+    medical_thermometer_state_t *state = (medical_thermometer_state_t *)context;
+
+    switch (event.event_type) {
+        case EVENT_ACTIVATE:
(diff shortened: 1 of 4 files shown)


movement/watch_faces/sensor/medical_thermometer_face.h
```diff
@@ -0,0 +1,71 @@
...
+ * This watch face shows the current temperature in degrees Celsius. Press and
+ * hold the alarm button to toggle between Celsius and Fahrenheit. Press and
+ * release the alarm button to start a "timer". The watch will sound an alarm
+ * when the temperature remains constant for at least 30 seconds and the
+ * temperature will stop updating until you press the alarm button. You can
+ * cancel the alarm by pressing the button again. If the temperature doesn't
+ * remain constant until the low energy timeout is reached, the alarm will stop.
+ * This measurement isn't super accurate, but if you put your watch under your
+ * shoulder when you feel sick, you will get an approximation of whether you're
+ * just feeling bad or it's serious.
```
matheusmoreira, line 42: We definitely need more data about the sensor watch thermistor's error margins. With error ≥ ±0.5 °C it could actually delay medical attention by giving users a false sense of normalcy.

movement/watch_faces/sensor/medical_thermometer_face.h
```diff
@@ -0,0 +1,71 @@
...
+#include "movement.h"
+
+/*
+ * MEDICAL THERMOMETER
+ *
+ * This watch face shows the current temperature in degrees Celsius. Press and
+ * hold the alarm button to toggle between Celsius and Fahrenheit. Press and
+ * release the alarm button to start a "timer". The watch will sound an alarm
+ * when the temperature remains constant for at least 30 seconds and the
+ * temperature will stop updating until you press the alarm button. You can
```
matheusmoreira, line 37: Sounds great. This is a familiar interface to anyone who has used a medical thermometer.

movement/watch_faces/sensor/medical_thermometer_face.h
```diff
@@ -0,0 +1,71 @@
...
+ * cancel the alarm by pressing the button again. If the temperature doesn't
+ * remain constant until the low energy timeout is reached, the alarm will stop.
```
matheusmoreira, line 39: Just the alarm? Will the face continue the temperature fix process in low energy mode? I ask because I use a ten second low energy timeout. After reading the code, to me it looks like it continues to read out temperatures even in low energy mode. Please confirm if that's you intention. Others may have concerns about power usage.

movement/watch_faces/sensor/medical_thermometer_face.c
```diff
@@ -0,0 +1,151 @@
...
+            if (watch_rtc_get_date_time().unit.second % 5 == 0) {
+                switch (state->mode) {
+                    case MODE_NORMAL:
+                        _medical_thermometer_face_update(settings->bit.use_imperial_units);
+                        break;
+                    case MODE_ALARM:
+                        for (size_t i = LAST_SIZE - 1; i > 0; i--) {
+                            state->last[i] = state->last[i - 1];
+                        }
+                        state->last[0] = roundf(_medical_thermometer_face_update(settings->bit.use_imperial_units) * 10.0f);
```
matheusmoreira, line 86: What is the impact of this rounding and multiplication on the accuracy of the measurements?

movement/watch_faces/sensor/medical_thermometer_face.h
Kistelini: I don't have any accurate thermometer at hand, but it's definitely not accurate enough to decide whether you need medical attention or not.

movement/watch_faces/sensor/medical_thermometer_face.h
Kistelini: It continues to read the temperature in low energy mode and updates every five minutes. That's the same behavior as the temperature readout face. The default minimum timeout for low energy mode is ten minutes, for ten seconds, this won't work.

movement/watch_faces/sensor/medical_thermometer_face.c
Kistelini: I'm rounding to the first decimal place and store the result in an integer, because you can't really compare floats in C.

joeycastillo:
The word “medical” gives me serious pause here, along with the usage description. Even with the disclaimer in the license, there are significant liabilities associated with suggesting that a device could diagnose any kind of medical condition, especially when I cannot guarantee the hardware is capable of such a thing. 

I think it’s cool that you’ve done this, but for liability reasons, it’s too great a risk to me personally for this to be part of the main repository.

matheusmoreira:
@joeycastillo I share your concern.

When I worked on the pulsometer I carefully avoided the word "medical" or any discussion of the accuracy whatsoever. I actually use it every day at work and it's awesome but the fact remains that without systematic testing no guarantees can be made from a product quality point of view.

Medical thermometers are simple devices [but are still Class II][1] medical devices according to the FDA. Unlike the [mercury clinical thermometers][2], the electronic ones are not exempt from 510(k) premarket notification and clearance. And this is just for the USA.

Just in case anyone is curious why a seemingly simple thermometer would be a regulated class II device: the reason is that all measurements have error, which creates uncertainty and therefore risk. Specifically, there is a risk of false positives and false negatives. A false positive is a feverish readout for a normal patient, and a false negative is a normal readout for a feverish patient.

Electronic medical and veterinary thermometer specifications I have evaluated in the past four years always had error margins ≤ ±0.5 ℃. Medical thermometer error usually hovered around ±0.2 ℃.

An error of ±0.2 ℃ looks low but it's actually pretty bad. Assuming a normal distribution and 95% confidence interval, if the device reads 37.5 ℃ then it means there's a 95% chance the _true_ body temperature lies within the [37.1, 37.9] ℃ interval. Even with a regulated, tested and calibrated medical thermometer, you cannot be reasonably certain the patient is feverish in that case. A readout of 38.0 ℃ is better: the confidence interval is [37.6, 38.4] ℃, you are 95% certain the patient's true body temperature is above the typical 37.5 ℃ threshold for fever.

An error of ±0.5 ℃ makes the device much harder to use for medical purposes. A readout of 38.0 ℃ would mean a confidence interval of [37.0, 39.0] ℃, putting that measurement in the same ambiguous category as the 37.5 ℃ ±0.2 ℃ readout. A value of 38.5 ℃ would be needed to be 95% certain the patient is feverish.

Every bit of error potentially delays treatment. This is literally a matter of life or death. [For example][3], criteria for sepsis include body temperature > 38.0 ℃, and early diagnosis and treatment predicts lower mortality.

To make any kind of claim whatsoever regarding the accuracy of the temperature readout, we would need to systematically measure it and compile enough statistics to establish the temperature's standard deviation and confidence interval. We'd also need to account for variability in the parts and manufacturing process.

Even without any medical device ambitions, determining these error margins would be a worthy effort. I'm sure the folks who implemented the watch's temperature compensation functionality would appreciate it. Still, it's unlikely to be reachable with available resources and manpower of an open source project.

[1]: https://www.accessdata.fda.gov/scripts/cdrh/cfdocs/cfpcd/classification.cfm?id=FLL
[2]: https://www.accessdata.fda.gov/scripts/cdrh/cfdocs/cfpcd/classification.cfm?id=FLK
[3]: https://www.ncbi.nlm.nih.gov/books/NBK598311/

Kistelini:
@joeycastillo I understand your concern. But I think the functionality is still useful, e.g. to receive a notification when you want to measure the room temperature after the watch has been taken off the wrist.
I pushed a new commit and removed all references to this being of medical use. I think the PR didn't update because it's closed. Would you please consider reopening it? Commit is here: https://github.com/joeycastillo/Sensor-Watch/commit/b51da0088588f777ebd8e2efc411a85163f8ad8a

joeycastillo:
Ah alas I can't seem to reopen it because the branch was force-pushed (this seems to cause the “reopen” button to become disabled). 

If you rename the branch, you should be able to open a new pull request! And I'd totally consider it, as long as the documentation in the watch face makes clear that it's not intended to diagnose, treat, cure, or prevent any disease.

PR #414, opened 2024-07-07
Added ability to make default time on the watch be PC clock's during …

voloved:
When running `make`, you can include `DATE=`, which will set the default time on the watch to the PC's.
Below are example outputs of the setting.

```
user@DESKTOP-UCQND0L:~/sensorWatch/Sensor-Watch/movement/make$ make COLOR=RED
git submodule update --init
size:
   text    data     bss     dec     hex filename
 109436    2676   11604  123716   1e344 build/watch.elf
 109436    2676   11604  123716   1e344 (TOTALS)
 
user@DESKTOP-UCQND0L:~/sensorWatch/Sensor-Watch/movement/make$ make COLOR=RED DATE=YEAR
Default year and timezone are set to 2024 EDT
git submodule update --init
size:
   text    data     bss     dec     hex filename
 109436    2676   11604  123716   1e344 build/watch.elf
 109436    2676   11604  123716   1e344 (TOTALS)
 
user@DESKTOP-UCQND0L:~/sensorWatch/Sensor-Watch/movement/make$ make COLOR=RED DATE=DAY
Default date set to Jul 07 2024 EDT
git submodule update --init
size:
   text    data     bss     dec     hex filename
 109436    2676   11604  123716   1e344 build/watch.elf
 109436    2676   11604  123716   1e344 (TOTALS)
 
user@DESKTOP-UCQND0L:~/sensorWatch/Sensor-Watch/movement/make$ make COLOR=RED DATE=MIN
Default time set to 12:45 on Jul 07 2024 EDT
git submodule update --init
size:
   text    data     bss     dec     hex filename
 109436    2676   11604  123716   1e344 build/watch.elf
 109436    2676   11604  123716   1e344 (TOTALS)
 
user@DESKTOP-UCQND0L:~/sensorWatch/Sensor-Watch/movement/make$ make COLOR=RED DATE=foo
../../make.mk:276: *** DATE must be YEAR, DAY, or MIN if used..  Stop.
```

[Context in Discord](https://discord.com/channels/969627287429459988/1029086310192918538/1259302478839545887)

### DIFF
--- a/make.mk
+++ b/make.mk
@@ -277,7 +277,7 @@ CFLAGS += -DINITIAL_MONTH=$(CURRENT_MONTH)
 CFLAGS += -DINITIAL_DAY=$(CURRENT_DAY)
 CFLAGS += -DINITIAL_HOUR=$(CURRENT_HOUR)
 CFLAGS += -DINITIAL_MINUTE=$(CURRENT_MINUTE)
-$(info Default time set to $(CURRENT_HOUR):$(CURRENT_MINUTE) on $(shell date +"%b") $(CURRENT_DAY) $(shell date +"%Y") $(shell date +%Z))
+$(info Default time set to $(CURRENT_HOUR):$(shell printf "%02d" $(CURRENT_MINUTE)) on $(shell date +"%b") $(CURRENT_DAY) $(shell date +"%Y") $(shell date +%Z))
 else
 $(error DATE must be YEAR, DAY, or MIN if used.)
 endif


movement/movement.c
```diff
@@ -384,6 +384,14 @@ void app_init(void) {
     movement_state.settings.bit.to_interval = MOVEMENT_DEFAULT_TIMEOUT_INTERVAL;
     movement_state.settings.bit.le_interval = MOVEMENT_DEFAULT_LOW_ENERGY_INTERVAL;
     movement_state.settings.bit.led_duration = MOVEMENT_DEFAULT_LED_DURATION;
+#ifdef MAKEFILE_TIMEZONE
+    for (int i = 0, count = sizeof(movement_timezone_offsets) / sizeof(movement_timezone_offsets[0]); i < count; i++) {
+        if (movement_timezone_offsets[i] == MAKEFILE_TIMEZONE) {
+            movement_state.settings.bit.time_zone = i;
+            break;
+        }
+    }
```
matheusmoreira, line 393: I feel like there should be a way to set this statically, without looping over the time zones at runtime. We need to figure out the index of the timezone. Perhaps the makefile can figure out or call a script that figures it out, then it can pass the index via preprocessor definition.

watch-library/hardware/main.c
```diff
@@ -69,7 +69,23 @@ int main(void) {
...
+#endif
+#ifdef MAKEFILE_CURR_DAY
+        date_time.unit.day = MAKEFILE_CURR_DAY;
+#endif
+#ifdef MAKEFILE_CURR_HOUR
+        date_time.unit.hour = MAKEFILE_CURR_HOUR;
+#endif
+#ifdef MAKEFILE_CURR_MINUTE
+        date_time.unit.minute = MAKEFILE_CURR_MINUTE;
+#endif
```
matheusmoreira, line 88: I would suggest defining static functions that return the initial time components and putting the `#ifdef`s in them. This avoids adding complexity to the initialization function. They'll just return constants so the compiler is likely to eliminate the functions when they are compiled.

```C
static uint16_t get_initial_year(void)
{
#ifdef MAKEFILE_CURR_YEAR
    return MAKEFILE_CURR_YEAR;
#else
    return 3;
#endif
}

// ...

date_time.unit.year = get_initial_year();
```

make.mk
```diff
@@ -242,3 +242,37 @@ endif
...
+CFLAGS += -DMAKEFILE_CURR_MONTH=$(CURRENT_MONTH)
+CFLAGS += -DMAKEFILE_CURR_DAY=$(CURRENT_DAY)
+$(info Default date set to $(shell date +"%b") $(CURRENT_DAY) $(shell date +"%Y") $(shell date +%Z))
+else ifeq ($(DATE), MIN)
+CFLAGS += -DMAKEFILE_TIMEZONE=$(TIMEZONE)
+CFLAGS += -DMAKEFILE_CURR_YEAR=$(CURRENT_YEAR)
+CFLAGS += -DMAKEFILE_CURR_MONTH=$(CURRENT_MONTH)
+CFLAGS += -DMAKEFILE_CURR_DAY=$(CURRENT_DAY)
+CFLAGS += -DMAKEFILE_CURR_HOUR=$(CURRENT_HOUR)
+CFLAGS += -DMAKEFILE_CURR_MINUTE=$(CURRENT_MINUTE)
```
matheusmoreira, line 273: I suggest not mentioning `MAKEFILE` in these variables. Users might add them to a header file and include it instead. The makefile therefore wouldn't be the only source of these values. See also PR #295 for example.

I suggest `INITIAL_TIMEZONE`, `INITIAL_YEAR`, and so on.

make.mk
```diff
@@ -242,3 +242,37 @@ endif
...
+#  YEAR = Sets the year and timezone to the PC's
+#  DAY = Sets the default time down to the day (year, month, day, timezone)
+#  MIN = Sets the default time down to the minute (year, month, day, timezone, hour, minute)
+ifdef DATE
+TIMEZONE := $(shell date +%z | awk '{print substr($$0, 1, 3) * 60 + substr($$0, 4, 2)}')
+CURRENT_YEAR := $(shell echo $$(($(shell date +"%Y") - 2020)))
+CURRENT_MONTH := $(shell date +"%-m")
+CURRENT_DAY := $(shell date +"%-d")
+CURRENT_HOUR := $(shell date +"%-H")
+CURRENT_MINUTE := $(shell date +"%-M")
```
matheusmoreira, line 256: This adds a direct dependency on GNU coreutils for building the firmware. I consider that to be perfectly OK but then again I've been running Linux for over a decade. I have no idea how this impacts Windows users. Maintainers might want to double check this.

watch-library/hardware/main.c
```diff
@@ -69,7 +114,11 @@ int main(void) {
     // if date/time register is 0 (power on reset state), default year to 2023.
     watch_date_time date_time = watch_rtc_get_date_time();
     if (date_time.reg == 0) {
-        date_time.unit.year = 3;
+        date_time.unit.year = get_initial_year();
+        date_time.unit.month = get_initial_month();
+        date_time.unit.day = get_inital_day();
+        date_time.unit.hour = get_inital_hour();
+        date_time.unit.minute = get_inital_minute();
```
mcguirepr89, line 121: Not sure if this is worth much, but it's been helpful for me. For Linux users who only add these 5 lines and don't change any other files, `sed` can do all the work (though it's a kind of hacky).

From `movement/make`:
```
sed -Ei "s/(unit.year = ).*(;)/\1$(( $(date +%-g) - 20 ))\2/;s/(unit.month = ).*(;)/\1$(date +%-m)\2/;s/(unit.day = ).*(;)/\1$(date +%-d)\2/;s/(unit.hour = ).*(;)/\1$(date +%-H)\2/;s/(unit.minute = ).*(;)/\1$(date +%-M)\2/" ../../watch-library/hardware/main.c && make COLOR=YOURCOLOR && make install COLOR=YOURCOLOR
```
I know it's not pretty, but it is handy.

matheusmoreira (approved):
Looks great!!

matheusmoreira:
This is a good PR but since movement 2.0 will bring about a new build system it seems pointless to merge this in right now. Let's wait and revisit this after the refactoring is done!

PR #416, opened 2024-07-08
Change 24 h with alarm

voloved:
The original Casio F-91W swaps between 24H mode by pressing the `ALARM` button.  
This PR does the same both on the `simple_clock_face` and the `clock_face`.  

`clock_face` will ignore this if `CLOCK_FACE_24H_ONLY` is set on make.

24h

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -377,7 +377,11 @@ void app_init(void) {
 
     memset(&movement_state, 0, sizeof(movement_state));
 
+#ifdef CLOCK_FACE_24H_ONLY
+    movement_state.settings.bit.clock_mode_24h = true;
+#else
     movement_state.settings.bit.clock_mode_24h = MOVEMENT_DEFAULT_24H_MODE;
+#endif
     movement_state.settings.bit.led_red_color = MOVEMENT_DEFAULT_RED_COLOR;
     movement_state.settings.bit.led_green_color = MOVEMENT_DEFAULT_GREEN_COLOR;
     movement_state.settings.bit.button_should_sound = MOVEMENT_DEFAULT_BUTTON_SOUND;
--- a/movement/watch_faces/clock/clock_face.c
+++ b/movement/watch_faces/clock/clock_face.c
@@ -105,6 +105,21 @@ static watch_date_time clock_24h_to_12h(watch_date_time date_time) {
     return date_time;
 }
 
+static void clock_toggle_24h_mode(movement_settings_t *settings, watch_date_time current) {
+    char buf[2 + 1];
+    settings->bit.clock_mode_24h = !settings->bit.clock_mode_24h;
+    if (clock_is_in_24h_mode(settings)) {
+        clock_indicate(WATCH_INDICATOR_PM, false);
+    }
+    else {
+        clock_indicate_pm(settings, current);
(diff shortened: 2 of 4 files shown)


voloved:
joeycastillo — Today at 4:55 PM
yeah i tend to agree, there are folks who feel.... strongly........ that their watch should never leave 24 hour mode. i'd venture that this feature on the original casio is an annoyance that folks just have to put up with

devolov — Today at 6:14 PM
I though that's what the CLOCK_FACE_24H_ONLY flag in make is for.
I saw clock_face uses it, but nothing else. I just added in a more global usage of that flag into the PR.
In my PR, if that flag is defined, pressing the alarm button doesn't toggle.
Also:
The preferences will no longer show the 24H toggle page
movement_state.settings.bit.clock_mode_24h will be set to true so any other items in the watch will also use 24H mode

joeycastillo — Today at 6:17 PM
It is, I'm just saying, I'm not wild about the UX of having that button toggle 24H mode in the simple watch face. It's a major change from how Sensor Watch has always worked, and on the one face that most everyone uses. Putting it in the settings menu was a choice, not an omission.

voloved:
joeycastillo — Today at 4:55 PM
yeah i tend to agree, there are folks who feel.... strongly........ that their watch should never leave 24 hour mode. i'd venture that this feature on the original casio is an annoyance that folks just have to put up with

devolov — Today at 6:14 PM
I though that's what the CLOCK_FACE_24H_ONLY flag in make is for.
I saw clock_face uses it, but nothing else. I just added in a more global usage of that flag into the PR.
In my PR, if that flag is defined, pressing the alarm button doesn't toggle.
Also:
The preferences will no longer show the 24H toggle page
movement_state.settings.bit.clock_mode_24h will be set to true so any other items in the watch will also use 24H mode

joeycastillo — Today at 6:17 PM
It is, I'm just saying, I'm not wild about the UX of having that button toggle 24H mode in the simple watch face. It's a major change from how Sensor Watch has always worked, and on the one face that most everyone uses. Putting it in the settings menu was a choice, not an omission.

matheusmoreira (commented):
It's my understanding that this probably won't be merged due to the watch's UX design concerns but still I'd like to say that the implementation looks good to me, especially the changes to `clock_face` which is the one I'm most familiar with. Thanks for your contribution.

voloved:
Thanks!
I actually made it so that is a toggle in the makefile, piggybacking off `CLOCK_FACE_24H_ONLY `.

```
# CLOCK_FACE_24H_ONLY = X
#  TOGGLE = Pressing the alarm button on the clock face toggles between 24H and 12H
#  YES = Always 24H mode
#  NO = 24H mode can be set in the settings
ifndef CLOCK_FACE_24H_ONLY
CLOCK_FACE_24H_ONLY := NO 
endif
ifeq ($(CLOCK_FACE_24H_ONLY), TOGGLE)
CFLAGS += -DCLOCK_FACE_24H_TOGGLE
else ifeq ($(CLOCK_FACE_24H_ONLY), YES)
CFLAGS += -DCLOCK_FACE_24H_ONLY
else ifneq ($(CLOCK_FACE_24H_ONLY), NO)
$(error CLOCK_FACE_24H_ONLY must be TOGGLE, YES, or NO if used.)
endif
```

PR #417, opened 2024-07-08
Clock face 24h only flag changes

voloved:
* Made the `CLOCK_FACE_24H_ONLY` flag more universal where it sets the `clock_mode_24h` bit on startup and doesn't allow it to be changed or seen in the preferences.
* Fixed typo of `PREFERENCES_FACE_NUM_PREFEFENCES`

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -377,7 +377,11 @@ void app_init(void) {
 
     memset(&movement_state, 0, sizeof(movement_state));
 
+#ifdef CLOCK_FACE_24H_ONLY
+    movement_state.settings.bit.clock_mode_24h = true;
+#else
     movement_state.settings.bit.clock_mode_24h = MOVEMENT_DEFAULT_24H_MODE;
+#endif
     movement_state.settings.bit.led_red_color = MOVEMENT_DEFAULT_RED_COLOR;
     movement_state.settings.bit.led_green_color = MOVEMENT_DEFAULT_GREEN_COLOR;
     movement_state.settings.bit.button_should_sound = MOVEMENT_DEFAULT_BUTTON_SOUND;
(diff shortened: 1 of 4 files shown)


movement/watch_faces/clock/clock_face.c
```diff
@@ -57,7 +53,6 @@ typedef struct {
 } clock_state_t;
 
 static bool clock_is_in_24h_mode(movement_settings_t *settings) {
-    if (CLOCK_FACE_24H_ONLY) { return true; }
```
matheusmoreira, line 60: In 24h only mode, this line reduces to:

    if (1) { return true; }

The purpose of this line is to allow the compiler to delete the function and replace it with `true` everywhere the code must branch into either 12h or 24h mode. The intention was to facillitate dead code elimination of the 12h mode code.

Can you confirm if this property still holds without the preprocessor definition?

movement/movement.c
```diff
@@ -377,7 +377,11 @@ void app_init(void) {
 
     memset(&movement_state, 0, sizeof(movement_state));
 
+#ifdef CLOCK_FACE_24H_ONLY
+    movement_state.settings.bit.clock_mode_24h = true;
+#else
     movement_state.settings.bit.clock_mode_24h = MOVEMENT_DEFAULT_24H_MODE;
+#endif
```
matheusmoreira, line 384: A suggestion: isolating this bit of complexity in a function could be good.

```C
void set_initial_clock_mode(void)
{
#ifdef CLOCK_FACE_24H_ONLY
// ...
#endif
}

set_initial_clock_mode();
```

matheusmoreira (approved):
Looks great to me!

PR #420, opened 2024-07-20
Added ability to add presets to the sunrise and sunset face

voloved:
I made it easy to cycle through other hardcoded coordinates in the sunrise senset face by pressing the `light` button.
If no presets are added except for the default, then it behaves exactly the same as it already does.
Adding more presets only will change the LED functionality, where a long-press causes the LED to start and clicking the LED button cycles. Holding the Alarm button when on a preset will default back to the default coordinates screen.

Example of adding other coordinates in my 'sunrise_sunset.h' file:
```
static const long_lat_presets_t longLatPresets[] =
{
    { .name = "  "},  // Default, the long and lat get replaced by what's set in the watch
//    { .name = "Ny", .latitude = 4072, .longitude = -7401 },  // New York City, NY
//    { .name = "LA", .latitude = 3405, .longitude = -11824 },  // Los Angeles, CA
//    { .name = "dE", .latitude = 4221, .longitude = -8305 },  // Detroit, MI
};
```

https://github.com/user-attachments/assets/6673804f-f45a-4989-8bc3-a085ed98d219

### DIFF
--- a/movement/watch_faces/complication/sunrise_sunset_face.c
+++ b/movement/watch_faces/complication/sunrise_sunset_face.c
@@ -399,7 +399,7 @@ bool sunrise_sunset_face_loop(movement_event_t event, movement_settings_t *setti
                 movement_request_tick_frequency(4);
                 _sunrise_sunset_face_update_settings_display(event, context);
             }
-            else{
+            else {
                 state->active_digit = 0;
                 state->page = 0;
                 _sunrise_sunset_face_update_location_register(state);
--- a/movement/watch_faces/complication/sunrise_sunset_face.c
+++ b/movement/watch_faces/complication/sunrise_sunset_face.c
@@ -372,7 +372,7 @@ bool sunrise_sunset_face_loop(movement_event_t event, movement_settings_t *setti
             else if (!state->page) movement_illuminate_led();
             break;
         case EVENT_LIGHT_BUTTON_UP:
-            if (state->page == 0) {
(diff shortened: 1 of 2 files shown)


movement/watch_faces/complication/sunrise_sunset_face.h
```diff
@@ -53,6 +53,7 @@ typedef struct {
     uint8_t active_digit;
     bool location_changed;
     watch_date_time rise_set_expires;
+    uint8_t longLatToUse;
```
theAlexes, line 56: please add new members to the end of the struct

movement/watch_faces/complication/sunrise_sunset_face.c
```diff
@@ -370,12 +388,23 @@ bool sunrise_sunset_face_loop(movement_event_t event, movement_settings_t *setti
...
+                _sunrise_sunset_face_update(settings, state);
+                break;
+            }
                 state->page++;
                 state->active_digit = 0;
                 watch_clear_display();
                 movement_request_tick_frequency(4);
                 _sunrise_sunset_face_update_settings_display(event, context);
             }
+            else{
```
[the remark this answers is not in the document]
voloved: done

movement/watch_faces/complication/sunrise_sunset_face.h
voloved: done

theAlexes (approved):

PR #421, opened 2024-07-20
Add secure mode to totp_face_lfs (secrets only in memory)

wryun:
Since the totp_lfs keeps things on flash unencrypted, if you're particularly paranoid this loads them from LFS then removes them from flash.

(WARNING: don't know enough about LFS to understand if they're properly cleared)

### DIFF
--- a/movement/watch_faces/complication/totp_face_lfs.c
+++ b/movement/watch_faces/complication/totp_face_lfs.c
@@ -41,6 +41,12 @@
 
 const char* TOTP_URI_START = "otpauth://totp/";
 
+/* The totp_record struct has two modes of operation:
+ *  - secmod = false: read from lfs (in which file_secret_offset
+ *    points to an offset in totp_uris.txt)
+ *  - secmod = true: read from memory
+ *    (in which case secret contains the decoded secret)
+ */
 struct totp_record {
     char label[2];
     hmac_alg algorithm;
@@ -232,7 +238,7 @@ static void totp_face_set_record(totp_lfs_state_t *totp_state, int i) {
     record = &totp_records[i];
 
     TOTP(
-        totp_face_lfs_get_file_secret(record),
+        (totp_state->secmod ? record->secret : totp_face_lfs_get_file_secret(record)),
         record->secret_size,
         record->period,
         record->algorithm
@@ -254,6 +260,7 @@ void totp_face_lfs_activate(movement_settings_t *settings, void *context) {
     }
 #endif
 
+    totp_state->secmod_request = false;
     totp_state->timestamp = watch_utility_date_time_to_unix_time(watch_rtc_get_date_time(), movement_timezone_offsets[settings->bit.time_zone] * 60);
     totp_face_set_record(totp_state, 0);
 }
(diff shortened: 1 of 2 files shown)


movement/watch_faces/complication/totp_face_lfs.c
```diff
@@ -279,15 +286,37 @@ static void totp_face_display(totp_lfs_state_t *totp_state) {
...
+        totp_records[i].secret = malloc(totp_records[i].secret_size);
+        if (totp_records[i].secret == NULL) {
+            num_totp_records = i - 1;
+            // We can't easily undo at this point, because we destroyed
+            // the offset info (secret is in a union). So let's just junk
+            // the remaining records.
+            break;
+        }
+        memcpy(totp_records[i].secret, current_secret, totp_records[i].secret_size);
+    }
```
matheusmoreira, line 303: How does this impact memory usage? Copying every secret into memory has caused resets in the past, and the fix was to read them into memory one at a time.

movement/watch_faces/complication/totp_face_lfs.c
```diff
@@ -279,15 +286,37 @@ static void totp_face_display(totp_lfs_state_t *totp_state) {
...
+
+    filesystem_rm(TOTP_FILE);
```
matheusmoreira, line 305: This is unlikely to be a secure erase operation. For performance reasons, file systems typically don't overwrite file contents when erasing, they just remove the links to the file from the file system data structures, causing the file to become unreachable and the memory to become available for new files. The data will not be overwritten until new data is written to the same physical location.

For this reason, it may be wise to overwrite the entire file with zeros before issuing the `rm` command to the file system.

This is adequate for spinning disks but may not be completely effective for flash storage: copies of the file may still linger on the storage medium if wear leveling is employed.

I do not know if the hardware supports a secure erase operation but it's probably a good idea to _also_ issue those commands if it does. Even if it does, it is unknown how effective it is, hardware manufacturers are known for screwing up features like this. The data might therefore still be recoverable despite all possible precautions.

matheusmoreira (commented):
There's no doubt that this commit raises the effort needed to recover data. With some changes the effort required might be increased even further. I would refrain from using the word "secure" though.

movement/watch_faces/complication/totp_face_lfs.c
wryun: Yes - there is a warning about this in the doco (i.e. that using secure mode is more likely to make you run out of memory, and whether it works well is a bit config dependent).

Personally, I've run 10 codes on my device without issue, but it obviously depends what other faces are loaded.

movement/watch_faces/complication/totp_face_lfs.c
wryun: :+1: 

Yes, as I mentioned at the top, don't know enough about LFS to understand exactly what will happen here. I wouldn't feel confident in zero-ing it out as a method because I haven't looked at LFS's internal structure (I mean, if it was the same length... but as you mention, LFS could have some built-in wear levelling that means it will write it in a separate block). At least there's no hardware level magic here we have to worry about!

I'll change the top of the PR to make these caveats clearer.

PR #424, opened 2024-07-31
Allow setting July 31st as today's date

edwardspeyer:
To allow a valid date to be set the `set_time_face` cycles through the valid days for a given month.  July was incorrectly encoded as having 30 days (it has 31!)


theAlexes:
whoops! thanks for the fix! :)

PR #428, opened 2024-08-03
Sunrise sunset presets 

voloved:
This is in response to my PR of  https://github.com/joeycastillo/Sensor-Watch/pull/420.
That PR is fine, but has a warning when building because there's an if/else statement that the compiler doesn't know will never run one of the statements, so it's been made more explicit.

Before the fix:
```
CC build/totp_face.o
CC build/totp_face_lfs.o
CC build/sunrise_sunset_face.o
CC build/countdown_face.o
CC build/sailing_face.o
../watch_faces/complication/pulsometer_face.c: In function 'pulsometer_display_title':
../watch_faces/complication/pulsometer_face.c:61:58: warning: unused parameter 'pulsometer' [-Wunused-parameter]
CC build/counter_face.o
CC build/blinky_face.o
CC build/moon_phase_face.o
CC build/accelerometer_data_acquisition_face.o
CC build/mars_time_face.o
CC build/orrery_face.o
CC build/astronomy_face.o
CC build/tomato_face.o
CC build/probability_face.o
CC build/wake_face.o
../watch_faces/complication/totp_face.c: In function 'totp_face_loop':
../watch_faces/complication/totp_face.c:208:47: warning: comparison of integer expressions of different signedness: 'int' and 'size_t' {aka 'unsigned int'} [-Wsign-compare]
CC build/frequency_correction_face.o
CC build/alarm_face.o
../watch_faces/complication/sunrise_sunset_face.c: In function '_sunrise_sunset_face_update':
../watch_faces/complication/sunrise_sunset_face.c:55:56: warning: array subscript 1 is above array bounds of 'const long_lat_presets_t[1]' {aka 'const struct <anonymous>[1]'} [-Warray-bounds]
In file included from ../watch_faces/complication/sunrise_sunset_face.c:31:
../watch_faces/complication/sunrise_sunset_face.h:80:33: note: while referencing 'longLatPresets'
CC build/ratemeter_face.o
CC build/interval_face.o
```

After the fix:
```
CC build/totp_face.o
CC build/totp_face_lfs.o
CC build/sunrise_sunset_face.o
CC build/countdown_face.o
CC build/sailing_face.o
CC build/counter_face.o
CC build/blinky_face.o
CC build/moon_phase_face.o
CC build/accelerometer_data_acquisition_face.o
CC build/mars_time_face.o
CC build/orrery_face.o
../watch_faces/complication/pulsometer_face.c: In function 'pulsometer_display_title':
../watch_faces/complication/pulsometer_face.c:61:58: warning: unused parameter 'pulsometer' [-Wunused-parameter]
CC build/astronomy_face.o
CC build/tomato_face.o
CC build/probability_face.o
../watch_faces/complication/totp_face.c: In function 'totp_face_loop':
../watch_faces/complication/totp_face.c:208:47: warning: comparison of integer expressions of different signedness: 'int' and 'size_t' {aka 'unsigned int'} [-Wsign-compare]
CC build/wake_face.o
CC build/frequency_correction_face.o
CC build/alarm_face.o
CC build/ratemeter_face.o
CC build/interval_face.o
```

### DIFF
--- a/movement/watch_faces/complication/sunrise_sunset_face.c
+++ b/movement/watch_faces/complication/sunrise_sunset_face.c
@@ -49,7 +49,7 @@ static void _sunrise_sunset_face_update(movement_settings_t *settings, sunrise_s
     double rise, set, minutes, seconds;
     bool show_next_match = false;
     movement_location_t movement_location;
-    if (state->longLatToUse == 0)
+    if (state->longLatToUse == 0 || _location_count <= 1)
         movement_location = (movement_location_t) watch_get_backup_data(1);
     else{
         movement_location.bit.latitude = longLatPresets[state->longLatToUse].latitude;
--- a/movement/watch_faces/complication/sunrise_sunset_face.c
+++ b/movement/watch_faces/complication/sunrise_sunset_face.c
@@ -359,7 +359,7 @@ bool sunrise_sunset_face_loop(movement_event_t event, movement_settings_t *setti
                     _sunrise_sunset_face_update_location_register(state);
                 }
                 _sunrise_sunset_face_update_settings_display(event, context);
-            } else if (_location_count == 1) {
+            } else if (_location_count <= 1) {
                 movement_illuminate_led();
             }
             if (state->page == 0) {
@@ -368,7 +368,7 @@ bool sunrise_sunset_face_loop(movement_event_t event, movement_settings_t *setti
             }
             break;
         case EVENT_LIGHT_LONG_PRESS:
-            if (_location_count == 1) break;
+            if (_location_count <= 1) break;
             else if (!state->page) movement_illuminate_led();
             break;
         case EVENT_LIGHT_BUTTON_UP:


matheusmoreira (approved):
Looks good. The `==` operator does not allow the compiler to infer that indexes are within range since it fails for values higher than 1. The `<=` coperator sufficiently limits the range of indexes to the expected values.

voloved:
From what I understood, the root cause was line 52.
if (state->longLatToUse == 0 || _location_count <= 1)
This is because state->longLatToUse == 0 will never be above 0 if _location_count is zero (defined elsewhere in the code).
The compiler doesn't know that it's blocked from this elsewhere, so being explicit here makes things better.
The _location_count <= 1 change was not related. It was to make it so the face allows the LED to turn on even if the array is completely empty (though it never should be)

voloved:
I do not think I can make the sunrise and sunset presets working perfectly due to DST.
The `sun_rise_set` returns the time in hours from UTC and they need to be converted into local time.
The local logic uses the timezone the user preset, but that requires manual update.
My commit below hardcodes the destinations' timezones, but that won't work during DST. We could check if we're within the days of DST, but what if we reference a location that doesn't use DST?

Here are the solutions I can see:
1. Always display the time in local time. So in my above screenshot, LA's sun does set at 10:10 my time (7:10 theirs).
2. Only allow locations in our timezone.
3. Only allow locations that are in our timezone OR don't use DST. Since 60% of the world doesn't use DST, this can be a large a mount. But it's still not perfect.
4. Implement a check on whether we're currently in DST weeks or not.
   - This will require additional look-up table to convert between the two, though it's been done in other PRs. But it does not fix the issue of us referencing locations that don't participate inDST.
5. Use our timezone and longitude for calculations. This causes 2 issues:
   1. the sunset calculation requires the longitude to be accurate for its calculations
   2. this will fail if our current timezone is in DST andthe target is not.
6. Drop this addition, and reimplement it when we have more comprehensive timezone logic, such as evq's utz library. 
   - That way, we can preset the timezone in the presets to a type of timezone, rather than a hardcoded offset. I tested it with New York's longitude and timezone and LA's latitude and got 7:16pm for sunset rather than 7:10. Not huge, but inaccurate. And also assumes both practice DST.

Sorry for this new bug, though I would not be surprised if I'm not the first person in this community to step on the rake that is DST....

matheusmoreira:
Don't sweat it, friend. I mean we're dealing with date and time here, basically one of the most hardest, insane things ever conceived by man. Let's give it our best try and if any issues come up we'll just deal with them.

There's a manual DST toggle PR which might help with this, PR #268, I'll get it merged into `next`.

PR #431, opened 2024-08-08
Hackwatch: July has 31 days

xtruan:
July has 31 days not 30. By chance I set up my sensor watch on July 31st and noticed this.


matheusmoreira (approved):
Thanks for fixing this!

PR #434, opened 2024-08-10
Can Set 29 on Feb on Leap Years Across All Set Faces

voloved:
Some of the watches have a message like this:
 > can't set to the 29th on a leap year. if it's february 29, set to 11:59 on the 28th.

This also changed July to having 31 days, rather than 30.
This change is a simple fix across all of those watches to allow Feb 29 to be set, but only on Leap Years.

### DIFF
--- a/apps/beats-time/app.c
+++ b/apps/beats-time/app.c
@@ -2,6 +2,7 @@
 #include <string.h>
 #include <math.h>
 #include "watch.h"
+#include "watch_utility.h"
 
 const int8_t UTC_OFFSET = 4; // set to your current UTC offset to see correct beats time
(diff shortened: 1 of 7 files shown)


matheusmoreira (approved):
Looks good. I especially like the improved functional organization via the `days_in_month` function. I think the more functions we have to hide gnarly time calculations, the better.

It looks perfectly functional as is but I do have a suggestion for future improvement. I see the following bit of logic repeat itself a lot:

```C
date_time.unit.year + WATCH_RTC_REFERENCE_YEAR
```

An additional helper function that offsets the time automatically could make the code easier to read and understand.

I'm also seeing a lot of repeated logic that rolls over days back to one, those could also benefit from a helper function.

PR #435, opened 2024-08-12
world_clock2_face.c bug fix

mcguirepr89:
Deleting this line removes an erroneous `movement_request_tick_frequency(1)` that prevents the `WORLD_CLOCK2_MODE_SETTINGS` `case` from keeping the tick frequency at `4` while setting the time zones. This allows the user to select their desired time zone with the intended smoothness gained with the higher tick frequency that is set on line 154


mcguirepr89:
Nevermind -- this is still a buggy face. Even after also addressing adding `movement_request_tick_frequency(1)` to line 363 and `movement_request_tick_frequency(4)` to line 366, the blinking is still not as smooth as it should be, and I'm unable to discern why.

PR #419, opened 2024-07-16
Endless runner

voloved:
# Endless Runner Face
This is a basic endless-runner, like the [Chrome Dino game](https://en.wikipedia.org/wiki/Dinosaur_Game).
* On the title screen, you can select a difficulty by long-pressing LIGHT or toggle sound by long-pressing ALARM.
* LED or ALARM are used to jump.
* High-score is displayed next to `HS:` on the title screen and can go up to 9999. It will reset at the beginning of every month.
* During a game, the current score is displayed on the top-right and cycle to 0 when it gets to 40.

# Difficulties
* **Baby** - Minimum 4 spaces between obstacles; Jumping is 3 frames; Half Speed
* **Easy** - Minimum 4 spaces between obstacles; Jumping is 3 frames; Normal Speed
* **Normal** - Minimum 4 spaces between obstacles; Jumping is 2 frames; Normal Speed
* **Hard** - Minimum 3 spaces between obstacles; Jumping is 2 frames; Normal Speed
* **Fuel** - The top-right counter displays fuel. So long as there's fuel, the runner can be above ground. The obstacles will come as waves. Each frame above ground subtracts fuel. Every frame on the ground adds one fuel. The max fuel is 30.
* **Fuel 1** - Same as `Fuel`, but if the fuel ever hits zero, it's game-over.

https://github.com/user-attachments/assets/896f53cb-e142-4a41-b739-8f3dec1e53db

PR #437, opened 2024-08-17
Debounce Logic

voloved:
# Debounce Logic
For Some, their watch buttons may trigger more than one press when clicked. This PR is used to compensate for that.  
What it does is ignore button presses after the first one for a set amount of ticks in `cb_fast_tick` (which uses a frequency of 128Hz).  

The amount of ticks to wait before registering a button is set by `DEBOUNCE_TICKS_DOWN` and `DEBOUNCE_TICKS_UP`.  Since the `cb_fast_tick` counts the ticks, and it's 128Hz, each tick is 7.8125ms.
* `DEBOUNCE_TICKS_DOWN`: How many ticks after pressing a button before another press of that button is registered.
* `DEBOUNCE_TICKS_UP`: How many ticks after depressing a button before another depress of that button is registered.

If both are set to 0, then the watch will behave the same as the stock firmware. Each of the three buttons has their own timer.

This does slow down button-press reactions, but it's only been seen as an issue on watch faces that require fast and precise presses, such as the [endless runner face](https://github.com/joeycastillo/Sensor-Watch/pull/419). In testing with 2 being the `DEBOUNCE_TICKS_DOWN` and `DEBOUNCE_TICKS_UP`, no other faces' usability was seen as affected.

`cb_fast_tick` is used for where the logic is handled to avoid introducing new timers when a button is pressed, but causes an issue, where that callback occurs asynchronous to when a button is pressed. If `DEBOUNCE_TICKS_DOWN` is set to 1 and the button is pressed when `cb_fast_tick` is 7ms into acting on its next callback, the debounce time is instead 0.8125 ms, not 7,8125. Due to this, setting the `DEBOUNCE_TICKS_DOWN` or `DEBOUNCE_TICKS_UP` to 1 is not recommended.

# Ignore Alarm Button After Sleep

This PR also includes logic that ignores the next time the Alarm button is depressed after waking from sleep. I have found it annoying that pressing the alarm button to wake the watch also triggers common behaviors on many faces at the same time.

# Diagram of the Debounce Method
Debounce Method

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -23,8 +23,8 @@
  */
 
 #define MOVEMENT_LONG_PRESS_TICKS 64
-#define DEBOUNCE_TICKS_DOWN 2  // In terms of *7.8125ms
-#define DEBOUNCE_TICKS_UP   2  // In terms of *7.8125ms
+#define DEBOUNCE_TICKS_DOWN 0  // In terms of *7.8125ms
+#define DEBOUNCE_TICKS_UP   0  // In terms of *7.8125ms
 
 #include <stdio.h>
 #include <string.h>
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -684,55 +684,29 @@ void cb_alarm_fired(void) {
     movement_state.needs_background_tasks_handled = true;
 }
 
+static void debounce_btn_press(uint8_t pin, uint8_t *debounce_ticks, bool *debounce_btn_trig, uint16_t *down_timestamp, void (*function)(bool)) {
+    if (*debounce_ticks > 0) (*debounce_ticks)--;
+    if (*debounce_btn_trig) {
+        bool pin_level = watch_get_pin_level(pin);
+        *debounce_btn_trig = false;
+        if (pin_level) {
+            function(pin_level);
+        }
+        else if (*debounce_ticks == 0) {
+            function(pin_level);
+            *debounce_ticks = DEBOUNCE_TICKS;
+        }
+        else {
+            *down_timestamp = 0;
+            _movement_disable_fast_tick_if_possible();
+        }
+    }
+}
+
 void cb_fast_tick(void) {
-    if (movement_state.debounce_ticks_light > 0) movement_state.debounce_ticks_light--;
-    if (movement_state.debounce_ticks_alarm > 0) movement_state.debounce_ticks_alarm--;
-    if (movement_state.debounce_ticks_mode > 0) movement_state.debounce_ticks_mode--;
-    if (movement_state.debounce_btn_trig_light) {
-        bool pin_level = watch_get_pin_level(BTN_LIGHT);
-        movement_state.debounce_btn_trig_light = false;
-        if (pin_level) {
-            light_btn_action(pin_level);
-        }
-        else if (movement_state.debounce_ticks_light == 0) {
-            light_btn_action(pin_level);
-            movement_state.debounce_ticks_light = DEBOUNCE_TICKS;
-        }
-        else {
-            movement_state.light_down_timestamp = 0;
-            _movement_disable_fast_tick_if_possible();
-        }
-    }
-    if (movement_state.debounce_btn_trig_alarm) {
-        bool pin_level = watch_get_pin_level(BTN_ALARM);
-        movement_state.debounce_btn_trig_alarm = false;
-        if (pin_level) {
-            alarm_btn_action(pin_level);
-        }
-        else if (movement_state.debounce_ticks_alarm == 0) {
-            alarm_btn_action(pin_level);
-            movement_state.debounce_ticks_alarm = DEBOUNCE_TICKS;
-        }
-        else {
-            movement_state.alarm_down_timestamp = 0;
-            _movement_disable_fast_tick_if_possible();
-        }
-    }
-    if (movement_state.debounce_btn_trig_mode) {
-        bool pin_level = watch_get_pin_level(BTN_MODE);
-        movement_state.debounce_btn_trig_mode = false;
-        if (pin_level) {
-            mode_btn_action(pin_level);
-        }
-        else if (movement_state.debounce_ticks_mode == 0) {
-            mode_btn_action(pin_level);
-            movement_state.debounce_ticks_mode = DEBOUNCE_TICKS;
-        }
-        else {
-            movement_state.mode_down_timestamp = 0;
-            _movement_disable_fast_tick_if_possible();
-        }
-    }
+    debounce_btn_press(BTN_LIGHT, &movement_state.debounce_ticks_light, &movement_state.debounce_btn_trig_light, &movement_state.light_down_timestamp, light_btn_action);
+    debounce_btn_press(BTN_ALARM, &movement_state.debounce_ticks_alarm, &movement_state.debounce_btn_trig_alarm, &movement_state.alarm_down_timestamp, alarm_btn_action);
+    debounce_btn_press(BTN_MODE, &movement_state.debounce_ticks_mode, &movement_state.debounce_btn_trig_mode, &movement_state.mode_down_timestamp, mode_btn_action);
     if (movement_state.debounce_ticks_light + movement_state.debounce_ticks_mode + movement_state.debounce_ticks_alarm  == 0)
         movement_state.fast_ticks++;
     if (movement_state.light_ticks > 0) movement_state.light_ticks--;
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -668,21 +668,19 @@ static void debounce_btn_press(uint8_t pin, uint8_t *debounce_ticks, uint16_t *d
     }
     else
         *down_timestamp = 0;
+    _movement_enable_fast_tick_if_needed();
 }
 
 void cb_light_btn_interrupt(void) {
     debounce_btn_press(BTN_LIGHT, &movement_state.debounce_ticks_light, &movement_state.light_down_timestamp, light_btn_action);
-    _movement_enable_fast_tick_if_needed();
 }
 
 void cb_mode_btn_interrupt(void) {
     debounce_btn_press(BTN_MODE, &movement_state.debounce_ticks_mode, &movement_state.mode_down_timestamp, mode_btn_action);
-    _movement_enable_fast_tick_if_needed();
 }
 
 void cb_alarm_btn_interrupt(void) {
     debounce_btn_press(BTN_ALARM, &movement_state.debounce_ticks_alarm, &movement_state.alarm_down_timestamp, alarm_btn_action);
-    _movement_enable_fast_tick_if_needed();
 }
 
 void cb_alarm_btn_extwake(void) {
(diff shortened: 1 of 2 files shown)


movement/movement.c
```diff
@@ -23,6 +23,8 @@
  */
 
 #define MOVEMENT_LONG_PRESS_TICKS 64
+#define DEBOUNCE_TICKS_DOWN 0  // In terms of *7.8125ms
+#define DEBOUNCE_TICKS_UP   0  // In terms of *7.8125ms
```
matheusmoreira, line 27: We should also add comments in the code itself showing how to arrive at this figure. It was calculated using the current 128 Hz interrupt but this could change in the future: I've seen high resolution timer work aiming at multiplexing the hardware timers and freeing up the 128 Hz tick for use by applications.

I think we also need detailed documentation here on why it's not recommended that the user set the values to 1. I'm not sure I understood the reason when I read the PR description.

movement/movement.c
```diff
@@ -462,6 +468,7 @@ void app_wake_from_standby(void) {
 
 static void _sleep_mode_app_loop(void) {
     movement_state.needs_wake = false;
+    movement_state.ignore_alarm_btn_after_sleep = true;
```
matheusmoreira, line 471: Does this need to be configurable? Is there a case where not ignoring erroneous bouncing button presses on watch wake up is advantageous?

I'd also advocate for it being a compile time configuration with preprocessor macros. This would allow the compiler to dead code eliminate all unused branches that depend on this boolean, reducing code size and memory usage.

movement/movement.c
```diff
@@ -627,29 +634,53 @@ static movement_event_type_t _figure_out_button_event(bool pin_level, movement_e
...
     _movement_reset_inactivity_countdown();
     event.event_type = _figure_out_button_event(pin_level, EVENT_LIGHT_BUTTON_DOWN, &movement_state.light_down_timestamp);
 }
 
-void cb_mode_btn_interrupt(void) {
-    bool pin_level = watch_get_pin_level(BTN_MODE);
+static void mode_btn_action(bool pin_level) { 
     _movement_reset_inactivity_countdown();
     event.event_type = _figure_out_button_event(pin_level, EVENT_MODE_BUTTON_DOWN, &movement_state.mode_down_timestamp);
 }
```
matheusmoreira, line 651: This is a great example of functional decomposition. It could be taken even further:

```C
static void btn_action(bool pin_level, int code, uint16_t *timestamp) {
    event.event_type = _figure_out_button_event(pin_level, code, timestamp);
}

static void light_btn_action(bool pin_level) {
    btn_action(pin_level, EVENT_LIGHT_BUTTON_DOWN, &movement_state.light_down_timestamp);
}

// mode_btn_action, implemented similarly to the above
```

movement/movement.c
```diff
@@ -627,29 +634,53 @@ static movement_event_type_t _figure_out_button_event(bool pin_level, movement_e
...
-void cb_alarm_btn_interrupt(void) {
-    bool pin_level = watch_get_pin_level(BTN_ALARM);
+static void alarm_btn_action(bool pin_level) {
     _movement_reset_inactivity_countdown();
-    event.event_type = _figure_out_button_event(pin_level, EVENT_ALARM_BUTTON_DOWN, &movement_state.alarm_down_timestamp);
+    uint8_t event_type = _figure_out_button_event(pin_level, EVENT_ALARM_BUTTON_DOWN, &movement_state.alarm_down_timestamp);
+    if  (movement_state.ignore_alarm_btn_after_sleep){
+        if (event_type == EVENT_ALARM_BUTTON_UP || event_type == EVENT_ALARM_LONG_UP) movement_state.ignore_alarm_btn_after_sleep = false;
+        return;
+    }
```
matheusmoreira, line 659: This branch could be dead code eliminated if the boolean variable is converted to compile time configuration with macros.

```C
    if  (MOVEMENT_IGNORE_ALARM_AFTER_SLEEP) {
        // ...
    }
```

If not configured to do that, it becomes:

```C
    if  (0) {
        // ...
    }
```

And the compiler eliminates the entire branch.

movement/movement.c
```diff
@@ -627,29 +634,53 @@ static movement_event_type_t _figure_out_button_event(bool pin_level, movement_e
...
+    debounce_btn_press(BTN_LIGHT, &movement_state.debounce_ticks_light, &movement_state.light_down_timestamp, light_btn_action);
+}
+
+void cb_mode_btn_interrupt(void) {
+    debounce_btn_press(BTN_MODE, &movement_state.debounce_ticks_mode, &movement_state.mode_down_timestamp, mode_btn_action);
+}
+
+void cb_alarm_btn_interrupt(void) {
+    debounce_btn_press(BTN_ALARM, &movement_state.debounce_ticks_alarm, &movement_state.alarm_down_timestamp, alarm_btn_action);
 }
```
matheusmoreira, line 684: Looks good!

movement/movement.c
```diff
@@ -662,7 +693,11 @@ void cb_alarm_fired(void) {
...
 
 void cb_fast_tick(void) {
-    movement_state.fast_ticks++;
+    if (movement_state.debounce_ticks_light > 0 && --movement_state.debounce_ticks_light == 0) _movement_disable_fast_tick_if_possible();
+    if (movement_state.debounce_ticks_alarm > 0 && --movement_state.debounce_ticks_alarm == 0) _movement_disable_fast_tick_if_possible();
+    if (movement_state.debounce_ticks_mode > 0 && --movement_state.debounce_ticks_mode == 0) _movement_disable_fast_tick_if_possible();
```
matheusmoreira, line 698: This bit of logic is ripe for being decomposed into a well named function.

movement/movement.h
```diff
@@ -270,6 +270,10 @@ typedef struct {
 
     // low energy mode countdown
     int32_t le_mode_ticks;
+    uint8_t debounce_ticks_light;
+    uint8_t debounce_ticks_alarm;
+    uint8_t debounce_ticks_mode;
+    bool ignore_alarm_btn_after_sleep;
```
matheusmoreira, line 276: This variable has the potential to be eliminated.

matheusmoreira (commented):
It's looking great and will be even greater when ready for the merge!

movement/movement.h
voloved: Which variable specidfically?

movement/movement.c
voloved: Is the size of a boolean a concern?
My worry is that more preprocessor macros will add more clutter.
I have found this logic needed as I allow toggling 24Hr mode with the ALARM button and waking the watch will always trigger that.
I cannot think of an event where a person would like to wake the watch and trigger the alarm button at the same time.
But I can ask the Discord.

movement/movement.c
voloved: Noted and change is made

movement/movement.c
voloved: I agree, if we want to make the variable a preprocessor step

movement/movement.c
voloved: Noted and modified

movement/movement.h
matheusmoreira: `ignore_alarm_btn_after_sleep`

movement/movement.c
matheusmoreira: I think it's a good idea to make it a preprocessor macro because I don't expect that users would want to change this setting at runtime. The discussions I see on discord appear to agree with this view point. It makes code more efficient and still allows people to revert to the previous behavior if needed.

movement/movement.c
matheusmoreira: > Is the size of a boolean a concern?

Perhaps. Since it is low power embedded software, I assume that memory is always a concern and am always trying to spot opportunities to reduce runtime state. I had to hot patch an out of memory bug in the TOTP face once...

I have analyzed the `movement_state_t` structure with and without the boolean:

```
WITH BOOLEAN

/*     44      |       1 */    uint8_t debounce_ticks_light;
/*     45      |       1 */    uint8_t debounce_ticks_alarm;
/*     46      |       1 */    uint8_t debounce_ticks_mode;
/*     47      |       1 */    _Bool ignore_alarm_btn_after_sleep;

                               /* total size (bytes):   56 */

WITHOUT BOOLEAN

/*     44      |       1 */    uint8_t debounce_ticks_light;
/*     45      |       1 */    uint8_t debounce_ticks_alarm;
/*     46      |       1 */    uint8_t debounce_ticks_mode;
/* XXX  1-byte hole      */

                               /* total size (bytes):   56 */
```

With the structure's current layout, the boolean simply occupies space that used to be padding. If we sort the structure members by size, we can reduce the structure's size to 48 bytes. This boolean just barely makes it into the structure without additional padding being added, it occupies the very last padding byte:

```
WITH BOOLEAN

/*     45      |       1 */    _Bool has_scheduled_background_task;
/*     46      |       1 */    _Bool needs_wake;
/*     47      |       1 */    _Bool ignore_alarm_btn_after_sleep;

                               /* total size (bytes):   48 */
WITHOUT BOOLEAN

/*     45      |       1 */    _Bool has_scheduled_background_task;
/*     46      |       1 */    _Bool needs_wake;
/* XXX  1-byte padding   */

                               /* total size (bytes):   48 */
```

So it's not causing any impact at the moment. The calculus could change if more state is added. For example, adding another boolean or byte sized variable to the structure would cause its size to increase to 52 bytes.

> My worry is that more preprocessor macros will add more clutter.

Do you think the macro adds clutter? I think they are similar enough in complexity to a boolean variable, perhaps even lower complexity since they don't exist at runtime.

I think if anything adds clutter its this configuration itself. Perhaps we could make the case for it being unnecessary and eliminate both the variable and the macro.

> I cannot think of an event where a person would like to wake the watch and trigger the alarm button at the same time.

Me neither! I think the behavior you proposed where the alarm input is ignored when the watch is woken up is the correct thing to do and that it shouldn't even need to be configured at all. Failing that, I think the next best option is making it a compile time toggle via preprocessor macro.

movement/movement.c
voloved: `ignore_alarm_btn_after_sleep` isn't asetting, it's a flag that goes high when we enter sleep and gets reset here.

matheusmoreira: Oh. My mistake, looks like I misunderstood its purpose. Please ignore all discussion regarding it.

movement/movement.c
```diff
@@ -627,29 +642,61 @@ static movement_event_type_t _figure_out_button_event(bool pin_level, movement_e
...
+    }
+    else
+        *down_timestamp = 0;
+}
+
+static void movement_disable_if_debounce_complete(void) {
+    if (movement_state.debounce_ticks_light > 0 && --movement_state.debounce_ticks_light == 0) _movement_disable_fast_tick_if_possible();
+    if (movement_state.debounce_ticks_alarm > 0 && --movement_state.debounce_ticks_alarm == 0) _movement_disable_fast_tick_if_possible();
+    if (movement_state.debounce_ticks_mode > 0 && --movement_state.debounce_ticks_mode == 0) _movement_disable_fast_tick_if_possible();
+}
```
matheusmoreira, line 688: Could be factored out further:

```C
static void disable_if_needed(uint8_t *ticks) {
    if (*ticks > 0 && --*ticks == 0)
        _movement_disable_fast_tick_if_possible();
}

static void movement_disable_if_debounce_complete(void) {
    disable_if_needed(&movement_state.debounce_ticks_light);
    disable_if_needed(&movement_state.debounce_ticks_alarm);
    disable_if_needed(&movement_state.debounce_ticks_mode);
}
```

matheusmoreira (approved):
I misunderstood the purpose of the boolean variable, so my suggestions to convert it to a macro do not apply. The code looks great for merging. Other than another minor suggestion I have no further comments. Looking forward to testing this vital feature on the watch, looks like people are already running it too. Thanks so much for improving this!

PR #439, opened 2024-08-18
make sure we don't miss our scheduled tasks

jpdbryant:
In testing a feature I'm adding, I was repeatedly pressing the "mode" button. I found that if I pressed the mode button just as a background task was supposed to fire, it didn't fire.

On the Discord, user tahnok suggested maybe the background task only checks for an exact time to fire, and indeed this seems to be the case. Changing this check from "is the correct time" to "is the correct time, or after" has fixed the errant behaviour I was seeing.


matheusmoreira (approved):
Good catch!

PR #440, opened 2024-08-18
Countdown fix

jpdbryant:
A couple of fixes to the Countdown face:

i) There's a race condition on calculating the delta - if we miss the moment when target_ts == now_ts, then we can get a negative delta, which we're storing as an unsigned int, so it overflows and things get weird. Instead, clamp it to 0.

ii) When the background task is called to indicate "time is up", it sounds the alarm and calls reset(), which clears the bell indicator. But it's a background task, so we shouldn't be changing the indicators - another face might be active. Instead, don't clear the indicator in reset(), clear it in draw().

### DIFF
--- a/movement/watch_faces/complication/countdown_face.c
+++ b/movement/watch_faces/complication/countdown_face.c
@@ -87,7 +87,10 @@ static void draw(countdown_state_t *state, uint8_t subsecond) {
 
     switch (state->mode) {
         case cd_running:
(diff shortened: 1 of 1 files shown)


matheusmoreira:
Seems correct. I see the `pause` function is also clearing indicators immediately after it cancels the background task. Doesn't it also need fixing?

matheusmoreira (approved):

PR #442, opened 2024-08-19
Update Devcontainer Dockerfile

nlordell:
This PR updates the .devcontainer Dockerfile which was no longer working. Attempting to build an image from the Dockerfile would result in an error:

```
E: The repository 'http://archive.ubuntu.com/ubuntu kinetic Release' does not have a Release file.
```

AFAIU, this is because the 22.04 release is no longer supported. I changed the base image to the current LTS release, which should remain supported until 2029.

In addition, the Dockerfile is also modified to install `emscripten` APT package. While this is not the recommented installation method from the Emscripten docs, it is much more convenient and works enough to build the simulator using the resulting Docker image.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,6 +1,4 @@
-FROM ubuntu:22.10
-
-# TODO: install emscripten (https://emscripten.org/docs/getting_started/downloads.html)
+FROM ubuntu:24.04
 
 # TODO: Clean this up once buildkit is supported gracefully in devcontainers
 # https://github.com/microsoft/vscode-remote-release/issues/1409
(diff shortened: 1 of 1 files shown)


matheusmoreira (approved):

matheusmoreira:
Looks good. I wonder if there's a way to use this devcontainer image from inside GitHub Actions. That would let us automatically verify that the project builds on it. I'm gonna look into this before merging.

matheusmoreira:
@nlordell

nlordell:
Sure, I don't mind looking into changing CI to run from the Devcontainer Docker container directly.

QQ, do you want to push the built image to a container registry (should speed up CI and allow others to `docker pull ...` to get a working Docker image and run a container capable of building the firmware)?

matheusmoreira:
Thanks! We should use the package registry since it will be faster and also allow developers to download the image for their own use. I think the repository owner needs to set up the credentials for it though. I've asked @joeycastillo on the discord.

In the meantime I'm gonna go ahead and merge this into the `next` branch. Please submit a second pull request later.

matheusmoreira:
Due to the new watch board's impending release and upcoming firmware refactoring, setting this up cannot be a priority at the moment. I'll open an issue to ensure it's not forgotten though. In the meantime the fix is already in main, thanks for contributing!

Issue #449, opened 2024-08-30
Cannot register two different callbacks with watch_register_extwake_callback


jpdbryant:
watch_register_extwake_callback does not wait for SYNCBUSY after updating the RTC config - when calling it twice in quick succession, it doesn't successfully register both callbacks. I have a patch for this, pull request forthcoming.

matheusmoreira:
PR #450 has been merged successfully. Thanks for improving this!

PR #450, opened 2024-08-30
watch_register_extwake_callback fix (re issue #449)

jpdbryant:
While working on something else, I found that if I call watch_register_extwake_callback twice in succession, to register two different callbacks, it doesn't register them both.

I think this is a race condition with the RTC - I have fixed by adding another wait for the SYNCBUSY bit. This also seems consistent with how this is done elsewhere in the code.

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -201,7 +201,7 @@ static void _movement_handle_scheduled_tasks(void) {
 
     for(uint8_t i = 0; i < MOVEMENT_NUM_FACES; i++) {
         if (scheduled_tasks[i].reg) {
-            if (scheduled_tasks[i].reg <= date_time.reg) {
+            if (scheduled_tasks[i].reg == date_time.reg) {
                 scheduled_tasks[i].reg = 0;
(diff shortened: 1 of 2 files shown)


matheusmoreira (approved):
Looks correct to me.

matheusmoreira:
I have already merged 20b4a32835d6b8aedda40cb49025cf3fb78beebe into `next` so I'll be dropping 897b55a6257c3150739bc83f4fb85812714f7d0c while merging this.

When making pull requests, base all work on top of the current `main` branch and just leave the merging to me! I'll make any necessary edits.

matheusmoreira:
Pull request did not close because 1549a5edc0d003b8f3002ef38cce7df7f8479b2b was not merged, cae5d8a33f5e7aa248b07e7a1b59fbc60b8c5a3b was merged instead. The hash of the commit changed because I rebased your pull request and dropped the two preceding commits. Your contribution is otherwise unchanged, however, and full attribution is kept. Thank you!

PR #451, opened 2024-08-30
Hourly chime disable at night

voloved:
# Hourly Chime Quiet Hours

This commit adds times when the chime won't play. Due to the limited size of `movement_settings_t`, this will only allow a few options for when to start and end chiming.  

Those options are:
 - First Hour that Chimes:
   - 6AM
   - 7AM
   - 10AM
   - Noon
   - Always
 - First Hour Without Chiming:
   - 8PM
   - 9PM
   - 10PM
   - Midnight
   - Always

## What preference Screen Looks Like
Start_chime
End_chime


PieterJanSterk:
I love this idea, is the code still functional? Will this be ported, implemented or even functional in the Second Movement firmware?

PieterJanSterk:
I found the code, over at second-movement/watch-faces/clock/clock_face.c line 280
For anyone who would like this functionality:

https://github.com/joeycastillo/second-movement/blob/918549a462f85e04c816402c927d6ba41dcbd594/watch-faces/clock/clock_face.c#L280

Thanks you so much, I always found this to be missing in het standard firmware. And indeed, I do not assume the setting itself would change that often. Does this also toggle the display marker for the hourly chime?

PR #456, opened 2024-08-31
Call set_time_face on long press of alarm.  Move alert toggle to light.

TonyHoyle:
Since we were talking about this :)

Instead of trying to merge the functionality into every watch face that might need it, I simply call the set_time_face directly on long press of alarm.

I've moved the alert functionality onto long press of light.

Tested on the simulator.. seems to work fine.  I did watch_face first then realized I was using simple_watch_face too but the changes are not large.

It would be nice if there was a formal way for faces to call complications/settings like this, but this works.


matheusmoreira:
Thank you for your contribution. The highest priority right now is the upcoming movement 2.0 refactor in preparation for the release of the new sensor watch pro board. Changes which touch the user interface changes will be considered after it is done.

> Instead of trying to merge the functionality into every watch face that might need it

> It would be nice if there was a formal way for faces to call complications/settings like this

You're correct about that. I too would like to improve this. I feel like there is a lot of duplicated functionality in between the numerous watch faces.

One of the purposes of movement's refactoring is to pay off tech debt. I plan to make contributions towards this end once the movement refactoring is under way. I encourage you to help out as well!

PR #457, opened 2024-08-31
Adding really long press for each button

mcguirepr89:
I don't expect this to be pulled into `main` anytime soon (if ever), but I wanted to flag it for the maintainers and folks who are paying close attention to this repo because I have found it increasingly useful to have another set of button events for each button for my own faces.

All this does is adds an "EVENT_*_REALLY_LONG_{PRESS,UP}" for each button in the movement framework.

Thanks again for everything and feel free to close this as soon as you'd like!

My best,
Patrick

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -1,7 +1,7 @@
 /*
  * MIT License
  *
- * Copyright (c) 2022 Joey Castillo
+ * Copyright (c) 2022 Joey Castillo (edits by Patrick McGuire)
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


matheusmoreira:
Thank you for your contribution! Movement will soon be refactored in preparation for the new sensor watch board release. For that reason, it was decided to postpone the merging of changes which touch the core of the framework. We hope to revisit this proposal after movement 2.0 is released.

mcguirepr89:
Ugh rebasing is a pain in the butt and I'm obviously not very good at it. I suppose we'll look at this again when the time is right. If things get too messy, I can always resubmit this as a clean PR with only the changes to `movement.c` and `movement.h` as I did originally instead of trying to rebase this PR.

mikemccabe:
Chiming in to say I think this is a great idea - especially the parity-with-Casio part!

That screen ('Sensor') would also be a great place to display some firmware ID strings: build date, version, git hash, builder-specified, etc.

PR #462, opened 2024-09-02
Change alarm to use long press to set

TonyHoyle:
Per conversation about making all the faces behave the same.

This puts the setting option for alarm watch face on long press of alarm, and moves the enable/disable alarm to the light long press.


matheusmoreira:
Thank you for your contribution. The highest priority right now is the upcoming movement 2.0 refactor in preparation for the release of the new sensor watch pro board. User interface changes will be considered after it is done.

PR #463, opened 2024-09-02
Require long press on alarm to modify prefs

TonyHoyle:
Per conversation about making all the faces behave the same.

This makes prefs behave like other faces, remaining in a read mode until a long press of alarm is used to set it.


joeycastillo:
Preferences is intended to be a settings-specific face, there’s no need for it to have a read-only mode.

PR #464, opened 2024-09-03
sleep light function

jpdbryant:
This enables the LIGHT button to wake the watch, if the watch has been physically modified to connect the LIGHT button to the A4 input. The watch cannot physically wake from the LIGHT input without this modification. However the A4 input does support external wake.

So, this patch adds an additional "external wake" callback on the A4 button.

The feature is configurable in movement_config.h by defining MOVEMENT_SLEEP_LIGHT_A4

The use-case is "checking the watch in the middle of the night". In order to preserve watch battery, the watch only wakes for 5 seconds. In order to accomplish this, there's some logic to disregard the LIGHT button for a few seconds after this special wake - otherwise, the "normal" LIGHT button action is triggered in addition to the A4 wake, which resets the low-energy countdown, which we don't want.

Room for improvement:

 * Perhaps this "don't respond" logic could be combined with the new debounce logic somehow?
 * Maybe the "don't respond" logic is just complicating things and it's okay to wake the watch up completely? (It does seem like potentially a lot of battery life though if you check the time a few times each night and it wakes up for an hour each time).
 * In movement.h there is no support for movement_config and so the additional struct field is not wrapped in in #ifdef

### DIFF
--- a/movement/movement.c
+++ b/movement/movement.c
@@ -192,6 +192,9 @@ void cb_mode_btn_interrupt(void);
 void cb_light_btn_interrupt(void);
 void cb_alarm_btn_interrupt(void);
 void cb_alarm_btn_extwake(void);
+#ifdef MOVEMENT_SLEEP_LIGHT_A4   
+void cb_a4_extwake(void);
+#endif
 void cb_alarm_fired(void);
 void cb_fast_tick(void);
 void cb_tick(void);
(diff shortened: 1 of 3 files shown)


matheusmoreira:
Thanks for your contribution. The movement framework will soon be refactored in preparation for a new sensor watch board. For this reason we decided to postpone evaluation of changes which modify movement's core functionality. We hope to revisit this PR after the refactor.

PR #465, opened 2024-09-04
add metal gear solid codec tune

neutralinsomniac:
only tested on simulator; would appreciate a hardware verification :)


matheusmoreira:

PR #466, opened 2024-09-04
Add harry potter signals

mcguirepr89:
This add a short and long Harry Potter signal


mcguirepr89:
Whoops!!! It also accidentally added all my customizations, which was not my intent.

PR #470, opened 2024-09-08
DST-Toggle

voloved:
This PR branches off of this PR: https://github.com/joeycastillo/Sensor-Watch/pull/268

This PR does 2 big things:

1. When DST starts or ends, it automatically rolls the time backwards or forwards.
2. Instead of modifying the timezone bit in settings, all functions that used to use the movement_timezone_offsets array now call a function that returns the offset based on if DST is currently occurring.  

So, this essentially makes the DST toggle an automatic one.

Video Shows how the toggle auto-changes the current timezone displayed and that changing the date outside of DST changes the timezone as well.

[ 2024-08-02_14-48-14.mp4 
](https://github.com/user-attachments/assets/dcd08e8a-06a9-48f8-885c-891260a7e2be)


matheusmoreira (approved):
Looks great!!

watch-library/shared/watch/watch_utility.c
```diff
@@ -83,6 +83,44 @@ uint8_t is_leap(uint16_t y)
 	return !(y%4) && ((y%100) || !(y%400));
 }
 
+uint8_t get_dst_status(watch_date_time date_time) {
```
harmonicoscillator, line 86: I think this function as currently written only applies to US DST rules. European DST dates are different: https://greenwichmeantime.com/daylight-saving-time/rules/eu/

This might need a toggle for US or EU DST rules in the settings? Or maybe a fully-flexible setting for the user to input what days/times to apply DST? None of these options are good but DST just sucks :(.

PR #473, opened 2024-09-10
Removed reliance on timezones for the countdown face.

voloved:
Our countdown face does not need to know what the current timezone is in order to work.
All logic can be done via GMT.


matheusmoreira:
This will probably have to be rewritten after the movement 2.0 refactor which will feature a new timezone API.

PR #477, opened 2024-09-19
allow setting alarm presets at compile time

mcguirepr89:
This allows a user to create preset alarms that are setup at compile time. This allows users to forego recreating their alarms anytime they re-compile their firmware. I've found it helpful and suspect there may be others who also find this helpful.

Alarm presets are set in `alarm_face.h` and look something like this:
```
static const alarm_setting_t alarm_presets[] = {
    { ALARM_DAY_WORKDAY, 6, 0, 5, 1, true }, //day, hour, minute, beeps, pitch, enabled
    { ALARM_DAY_WORKDAY, 16, 30, 5, 1, true },
};
```


joeycastillo:
Hi and thanks for this pull request! Pull requests against this repository are on hold during a refactor of Movement to support new hardware. Still I think this is a promising addition and I'd like to look at it at a later date. Sorry if it doesn't get attention right away!

PR #479, opened 2024-09-28
Add Chirpy support to databank

johanoskarsson:
I just backed the Sensor Watch Pro campagin, so I figured I'd check out the code a bit. This PR adds Chirp support to the databank to get the data back out of the watch again.

I closely followed the chirpy demo code. More comments inline.

However: one issue remains. Since the light and alarm buttons are already used in both short press and long press I'm not sure what I should be using to trigger the chirping. One idea was simultaneous presses of the light and alarm buttons? I notice there is no specific event for this. I could probably work around that but I figured I'd ask for advice.

### DIFF
--- a/movement/watch_faces/complication/databank_face.c
+++ b/movement/watch_faces/complication/databank_face.c
@@ -27,6 +27,7 @@
 #include "databank_face.h"
 #include "watch.h"
 #include "watch_private_display.h"
+#include "chirpy_tx.h"
 
 const char *pi_data[] = {
     "PI", "314159265358979323846264338327950288419716939937510582097494459230781640628620899862803482534211706798214808651328230664709384460955058223172535940812848111745028410270193852110555964462294895493038196442",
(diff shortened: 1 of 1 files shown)


movement/watch_faces/complication/databank_face.c
```diff
@@ -40,12 +41,99 @@ const char *pi_data[] = {
 
 const int databank_num_pages = (sizeof(pi_data) / sizeof(char*) / 2);
 
+typedef enum {
+    DATABANK_DISPLAY = 0,
+    DATABANK_CHIRPING,
```
johanoskarsson, line 46: The chirping mode changes the ticks to speed up etc.

movement/watch_faces/complication/databank_face.c
```diff
@@ -40,12 +41,99 @@ const char *pi_data[] = {
...
+        _databank_quit_chirping();
+        return;
+    }
+    uint16_t period = chirpy_get_tone_period(tone);
+    watch_set_buzzer_period(period);
+    watch_set_buzzer_on();
+}
+
+// Which byte are we currently processing?
+static uint16_t curr_data_ix;
```
johanoskarsson, line 85: In the chirpy demo code we also have data length and pointer to data here. But I'm instead using just whatever the currently displayed data is.

movement/watch_faces/complication/databank_face.c
```diff
@@ -40,12 +41,99 @@ const char *pi_data[] = {
...
+
+static uint8_t _databank_get_next_byte(uint8_t *next_byte) {
+    // We always send out the currently displayed dataset
+    int curr_pos = databank_state.databank_page * 2 + 1;
```
johanoskarsson, line 89: I'll be honest I don't quite understand the pagination used into the pi_data array, but this seems to work.

PR #483, opened 2024-10-16
Add alternate sunrise sunset face

mcguirepr89:
This is just an alternate face to the sunrise sunset face that adds in civil and nautical twilight. There was also a lot of code refactoring to make it make more sense to me as I tinkered with this.

PS
I'm very excited for movement 2.0

PPS
I'm also planning to add an EVENT_ALARM_LONG_PRESS eventually that will allow the user to set these different solar events as alarms (to enjoy a sunrise/sunset or to be otherwise notified when the outside darkness has reached the civil twilight threshold).

https://github.com/user-attachments/assets/4c77cb11-04b5-46c6-ba0f-592fae39e368

For this video, I set the clock to really early to show the solar events for an entire day. Then I reset the time to show that when the solar event has passed for the day, the event gets put at the bottom of the loop, so you can see the time event times for the next day for any event that has passed.

### DIFF
--- a/movement/watch_faces/complication/sunrise_sunset_alt_face.c
+++ b/movement/watch_faces/complication/sunrise_sunset_alt_face.c
@@ -77,10 +77,11 @@ static void check_and_update_sunriset(double event_type, watch_date_time *event_
     }
 }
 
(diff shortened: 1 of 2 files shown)


mcguirepr89:
I just realized that my PR replaced the original -- let me fix that!

PR #523, opened 2025-06-04
Update discgolf_face.c

koba0908:
It seems that the `courses` variable should be defined in `discgolf.h` rather than in `watch.h`, so the comment in `discgolf.c` appears to be misplaced by one line.

### DIFF
--- a/movement/watch_faces/complication/discgolf_face.c
+++ b/movement/watch_faces/complication/discgolf_face.c
@@ -1,7 +1,7 @@
 #include <stdlib.h>
 #include <string.h>
-#include "discgolf_face.h"
-#include "watch.h"          // Remember to change number of courses in this file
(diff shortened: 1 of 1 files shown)


movement/watch_faces/complication/discgolf_face.c
```diff
@@ -1,7 +1,7 @@
 #include <stdlib.h>
 #include <string.h>
-#include "discgolf_face.h"
-#include "watch.h"          // Remember to change number of courses in this file
+#include "discgolf_face.h"　　　　// Remember to change number of courses in this file
```
kachick, line 3:
```suggestion
#include "discgolf_face.h"    // Remember to change number of courses in this file
```

It seems there are non-ASCII characters. Is this intentional?

